PR #2, opened 2016-06-16
API Updates for v1.1-dev (upgrade path and https, 1/3)

andrewbonney:
These changes cover:

- Clarifying the upgrade path and explicitly indicating when HTTPS is supported
  - Add DNS-SD TXT records for api_ver and api_proto
  - Add matching version and protocol data to the Node API /self resource
  - Add ‘secure’ boolean attribute to websocket subscriptions and optional support for Upgrade header to API endpoints
  - Prefer the DNS SRV record priority mechanism rather than TXT ‘pri’ record


andrewbonney:

PR #3, opened 2016-06-16
API Updates for v1.1-dev (advanced queries and paging, 2/3)

andrewbonney:
These changes cover:

- Enabling advanced queries and paging
  - Adds an optional paging mechanism to the Node and Query API
  - Adds an optional advanced query mechanism to the Query API, supplementing the existing basic functionality


andrewbonney:

PR #4, opened 2016-06-16
API Updates for v1.1-dev (minor tweaks and additions, 3/3)

andrewbonney:
These changes cover:

- Add a ‘controls’ attribute to Devices, permitting control endpoint exposure in the same style as Node ‘services’
- Add a device_id keys to Flow attributes, enabling Nodes which do not act as a Source (e.g. encoders, decoders, scalers)
- Permit ‘null’ flow_id against a Sender in order to allow advertisement before internal device routing has occurred

### DIFF
--- a/APIs/schemas/device.json
+++ b/APIs/schemas/device.json
@@ -10,7 +10,8 @@
     "type",
     "node_id",
     "senders",
-    "receivers"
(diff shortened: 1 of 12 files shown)


andrewbonney:

PR #27, opened 2017-10-03
schemas: restrict formatting of interface port ID and chassis ID

andrewbonney:
Schema validation modification to restrict Port IDs to be MAC addresses, and suggest formatting for Chassis IDs. Intended to match the representation agreed within the Network Control API.

### DIFF
--- a/APIs/schemas/node.json
+++ b/APIs/schemas/node.json
@@ -106,7 +106,7 @@
           }
         },
         "interfaces": {
-          "description":"Network interfaces made available to devices owned by this Node",
+          "description":"Network interfaces made available to devices owned by this Node. Port IDs and Chassis IDs are used to inform topology discovery via IS-06, and require that interfaces implement ARP at a minimum, and ideally LLDP.",
           "type": "array",
           "items": {
             "type": "object",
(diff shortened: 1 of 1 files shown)


thosil:
Hi,
is there a reason restricting only to the format with dashes? Linux uses colons, Cisco and Arista dots (ex: 0123.4567.89AB).
Don't you think it would be nice to accept these formats:

- 01-23-45-67-89-AB
- 01:23:45:67:89:AB
- 0123.4567.89AB
- 0123456789AB

andrewbonney:
Hi Thomas,
This constraint was added after discussion with the network control API group. This stems from the need to compare values between this IS-04 and IS-06 APIs in order to match up endpoints, and aims to ease interoperability by ensuring that a simple string comparison will always work. There was a danger that without this, a naive implementation may simply perform a string comparison which would work with a certain set of vendors' equipment, but then wouldn't work with others.

An alternative would have been to clearly specify the LLDP subtype alongside these fields (indicating a MAC address was in use), but at the time there was a desire to permit a fallback to endpoint identification via gratuitous ARP. This would have used a MAC address too, but advertisement of an LLDP subtype in this case may have caused confusion where LLDP simply wasn't implemented.

In short, whilst this may require a simple translation to format data from the API server's perspective, it should make the job of the API client simpler and less error-prone. Additionally it makes schema validation far more straightforward.

I hope that helps.

PR #40, opened 2018-02-27
Remove erroneous use of EmptySchema used to indicate empty request body

### LINKED ISSUE #36: Use of EmptySchema in GET/DELETE request specification is incorrect
Noted by Gareth Sylvester-Bradley:

The 'EmptySchema' used in the APIs is intended to indicate that the request body can be empty, but in reality if you run it through a validator it requires an empty object to be present '{}'.

This should be replaced with one of the following in the RAML to fix this. It doesn't appear to be possible to fix this correctly using JSON schema directly:

1) Currently preferred: Body parameter removed for requests with no request payload
```
get:
  description: List of paths available from this API
  responses:
    200:
```

2) Body set explicitly to null. Note that when rendered as HTML this indicates that the request body must still have a content type of application/json which isn't mandatory without a request body present.
```
get:
  description: List of paths available from this API
  body: null
  responses:
    200:
```

3) Similar to 2, but using !! to avoid 'null' being interpreted as a literal. This usage isn't 100% clear from the RAML 0.8 spec, but does still appear to validate.
```
get:
  description: List of paths available from this API
  body: !!null
  responses:
    200:
```

andrewbonney:
Fixes #36


garethsb (approved):
Thanks, Andrew. Definitely looks like "the simplest thing that could possibly work"! Since this fixes a mistake should this also be merged to v1.0.x and v1.1.x?

andrewbonney:
Yes, I'm intending to backport documentation and bug pull requests wherever possible once the initial merge is done. If there are any which are deemed too contentious and aren't clearly enough a bug (ie. could be seen as a new feature) these may actually have to go into v1.3-dev to avoid any breaking changes.

PR #43, opened 2018-02-27
Add explicit OPTIONS support previously covered in docs

### LINKED ISSUE #16: OPTIONS support should be explicitly called out in the spec
When performing write requests (PUT/POST/DELETE etc) from a web browser a CORS pre-flight check will be performed to ensure that the client is permitted to make the request. This involves an OPTIONS request with headers similar to those described here: https://github.com/AMWA-TV/nmos-discovery-registration/blob/master/docs/2.3.%20APIs%20-%20Server%20Side%20Implementation%20Notes.md#cross-origin-resource-sharing-cors

At present the requirement to support OPTIONS is only called out here and whilst it has been tested at workshops it is a common stumbling block as it isn't explicitly in the RAML. OPTIONS support should be explicitly called out in all RAML which uses these sorts of methods.

andrewbonney:
Fixes #16

### DIFF
--- a/APIs/NodeAPI.raml
+++ b/APIs/NodeAPI.raml
@@ -244,6 +244,12 @@ documentation:
           body:
             schema: ErrorSchema
     /target:
+      options:
+        description: A pre-flight check generally used for Cross-Origin Resource Sharing (CORS) purposes
+        responses:
+          200:
+          403:
+          404:
       put:
         description: Request a change to a Receiver's subscription (deprecated)
         body:
--- a/APIs/QueryAPI.raml
+++ b/APIs/QueryAPI.raml
@@ -442,6 +442,11 @@ documentation:
             schema: ErrorSchema
 /subscriptions:
   displayName: Subscriptions
+  options:
+    description: A pre-flight check generally used for Cross-Origin Resource Sharing (CORS) purposes
+    responses:
+      200:
+      403:
   post:
     description: Create a websocket subscription to an API resource
     body:
@@ -488,6 +493,12 @@ documentation:
           description: Returned when the requested Subscription ID does not exist
           body:
             schema: ErrorSchema
+    options:
+      description: A pre-flight check generally used for Cross-Origin Resource Sharing (CORS) purposes
+      responses:
+        200:
+        403:
+        404:
     delete:
       description: Delete a single subscription
       body:
--- a/APIs/RegistrationAPI.raml
+++ b/APIs/RegistrationAPI.raml
@@ -54,6 +54,11 @@ documentation:
           schema: !include schemas/registrationapi-base.json
 /resource:
   displayName: Resource
+  options:
+    description: A pre-flight check generally used for Cross-Origin Resource Sharing (CORS) purposes
+    responses:
+      200:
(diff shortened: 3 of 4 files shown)


garethsb (changes requested):
I believe CORS preflight is also required if there are any custom request headers, so e.g. also on GET requests on the Query API. Is it possible in the RAML to describe this generally, like the default mediaType or using the traits mechanism somehow?
Also, when is a 403 Forbidden response to OPTIONS allowed?

andrewbonney:
I'm not sure on the general case. I'll take a further look, but was at least trying to cover the common error case which has been seen at past workshops.

The 403 I'd picked up on as a possibility for a response to the OPTIONS when the 'Access-Control-Request-X' don't match with the 'Access-Control-Response-X' but I'm failing to remember if this was added prior to or having found a suitable reference. The W3C spec doesn't seem completely clear in this regard at first glance so I'll investigate further and modify if required.

andrewbonney:
I've had a bit more of a read of the CORS spec. I'd suggest that for GET requests performed against any of these APIs it isn't necessary for a client to set any of the CORS-forbidden request headers. In the case where we have any kind of request body (with an application/json Content-Type), or use a modification request such as a DELETE (with no body) the CORS spec explicitly requires support for the pre-flight OPTIONS request.

For that reason my suggestion would be to include OPTIONS explicitly in the spec for these modification requests with an explicit requirement, but perhaps to add some extra guidance that server-side implementations MAY also support OPTIONS for all other methods/resources just in case a client chooses to set a forbidden header for any reason. Do you think that would be sufficient in order to cover the mandatory stuff but avoid clouding the RAML with OPTIONS/HEAD methods everywhere?

On the 403s: The W3C spec isn't totally clear on use of 403s as it's very much talking from the client's perspective, but whether you return that or a 200 which still indicates failures to match the sharing check then some form of 'network error steps' are ultimately performed which terminates the request (https://www.w3.org/TR/cors/#cross-origin-request-with-preflight-0). I'd suggest that the 403 just gives more power to the server to explicitly indicate a rejection rather than relying upon the client's 'resource sharing checks'.

There's an example of this at https://developer.mozilla.org/en-US/docs/Web/HTTP/Server-Side_Access_Control under "Preflighted requests". Where an OPTIONS request has been made using an Origin which isn't permitted, a 403 is returned. I guess technically the 403 should also be a valid response for all other API resources in this case (GETs included) as if these come from forbidden origins you'd expect the same response, but again as-per the W3C spec you could just return 2XXs and rely on the client to correctly perform a sharing check against the response headers. Let me know what you think - it seems like another case where this spec could become clouded in a long list of potential HTTP response codes for every resource.

garethsb:
Thanks, @andrewbonney. Of course, I was thinking of all the _response_ headers in the Query API. Sorry for the noise. In which case, I think expanding the documentation about CORS/OPTIONS slightly as you suggest, but in the RAML only calling out the OPTIONS request for the POST/PATCH/DELETE(/PUT) endpoints, is the right answer. Then just something for us to remember if we ever consider adding any _request_ headers in the future.

PR #44, opened 2018-02-27
docs: add note on referential integrity in device arrays

### LINKED ISSUE #24: Improve documentation around registration of the deprecated Device 'senders' and 'receivers' arrays
The 'senders' and 'receivers' arrays in a Device have been deprecated, but will continue to be present until v2.0.

Whilst clients' use of these fields is discouraged as a result, implementations have been seen which exhibit two behaviours. First first involves registering the Device with no Senders/Receivers in the arrays, then registering the Senders and Receivers before issuing a re-registration (update) of the Device with the arrays changed to reflect this. The second involves registering the Device with the arrays pre-populated before the Senders and Receivers are registered into the system.

The following text has been suggested as an addition to avoid clients attempting to request Senders/Receivers which do not yet exist in the system:
1.	Devices MAY be registered with UUIDs of the Senders and Receivers attached to the Device, in the (deprecated) "senders" and "receivers" array properties, even if those Senders/Receivers have not yet been registered themselves.
2.	Senders and Receivers attached to Devices MAY be omitted from the Device's (deprecated) "senders" and "receivers" array properties. Consumers of the Node/Query APIs SHOULD instead discover attached Senders/Receivers by filtering Senders and Receivers on "device_id".

andrewbonney:
Fixes #24


garethsb (approved):
Thanks. 👍

PR #45, opened 2018-02-27
examples: add namespaces and remove ports for ws_href examples

### LINKED ISSUE #33: Websocket example URIs
ws_href property in [this example](https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.2-dev/examples/queryapi-v1.1-subscriptionid-get-200.json) and similar should include the API namespace. In this example it should read ws://172.29.80.52:8870/x-nmos/query/v1.2/ws/?uid=7c903667-7113-4a8f-8865-1c63f9070a9e instead of ws://172.29.80.52:8870/ws/?uid=7c903667-7113-4a8f-8865-1c63f9070a9e.

As far as I can see this affects the following examples:
https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.2-dev/examples/queryapi-v1.1-subscriptionid-get-200.json
https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.2-dev/examples/queryapi-v1.1-subscriptions-get-200.json
https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.2-dev/examples/queryapi-v1.1-subscriptions-post-200.json

andrewbonney:
Fixes #33

### DIFF
--- a/examples/queryapi-v1.1-subscriptionid-get-200.json
+++ b/examples/queryapi-v1.1-subscriptionid-get-200.json
@@ -4,6 +4,6 @@
     "params": {},
     "persist": false,
     "secure": false,
-    "ws_href": "ws://172.29.80.52:8870/ws/?uid=7c903667-7113-4a8f-8865-1c63f9070a9e",
+    "ws_href": "ws://172.29.80.52/x-nmos/query/v1.2/ws/?uid=7c903667-7113-4a8f-8865-1c63f9070a9e",
     "id": "7c903667-7113-4a8f-8865-1c63f9070a9e"
 }
--- a/examples/queryapi-v1.1-subscriptions-get-200.json
+++ b/examples/queryapi-v1.1-subscriptions-get-200.json
@@ -5,7 +5,7 @@
         "params": {},
         "persist": false,
         "secure": false,
-        "ws_href": "ws://172.29.80.52:8870/ws/?uid=6a52dbd5-a737-4c4e-823f-909ade8f8bf4",
+        "ws_href": "ws://172.29.80.52/x-nmos/query/v1.2/ws/?uid=6a52dbd5-a737-4c4e-823f-909ade8f8bf4",
         "id": "6a52dbd5-a737-4c4e-823f-909ade8f8bf4"
     },
     {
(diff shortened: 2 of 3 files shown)


garethsb (changes requested):
Andrew, what's your rationale for using the "x-nmos" namespace here? This would seem to encourage vendors that they can put their own extensions under it?
Also, why remove the port? Although the default port for ws/wss is 80/443 same as http/https, wouldn't it be clearer to specify in the example?

andrewbonney:
I'm not sure I follow your first point. It is perfectly valid for a vendor to implement something else at a path of ws://<ip address>/another-namespace/ for example. The aim is just to bring the examples into line with all other paths in use for the APIs.

I can add the port back (as 80 or 443 for example if desired), but the reason for removing it was given that it indicated '8870' which was a port previously used by the reference implementation. We had seen incorrect interpretations which assumed you must use that port number.

These are simply intended to be examples so it doesn't matter from the perspective of the specification what URL is returned to a user for the websocket path, as long as it functions as expected.

garethsb:
I think using "x-nmos/query/v1.2" has the same problem with implementors making assumptions as specifying '8870' did.

I remember some previous discussion that having example URLs like:

http://{host}:{port}/x-nmos/query/v1.2/vendor-specific

or:

http://{host}:{port}/x-nmos/query/v1.2/senders/{senderId}/vendor-specific

was likely to result in implementors using such URLs, with the result that clients will assume those *are* part of the specification. IMHO, not good for interop, or future extensions to the specification.

For that matter, implementors are going to make assumptions based on the fact that the example uses a query parameter to hold the unique id, or even that the URL contains the same unique id as the "id" property at all.

I'd almost rather see e.g. "ws://host:12345/example/path/one", etc. in the examples.

andrewbonney:
Ok I see what you mean. The use of vendor-specific paths under the x-nmos namespace is certainly not something that's encouraged, even if it hasn't explicitly been called out as forbidden in the spec. If I'm understanding correctly the issue here is that as we don't specify the path '/ws' in the RAML it is confusing whether this is a specified path which has been omitted, and if not whether it means you can create whatever additional paths you would like.

Happy to do away with this change altogether if preferred. It came about because of https://github.com/bbc/nmos-query/issues/3, but perhaps a better approach would be to note somewhere that the URL you are passed back in the ws_href attribute will not follow any standard pattern.

peterbrightwell:
I would say definitely don't put vendor-specific things in the x-nmos tree (similar to urn:x-nmos).  We should consider making that formal.

PR #46, opened 2018-02-27
docs: indicate more strongly use of POSTs to begin websocket connections

### LINKED ISSUE #26: Clarify that websocket subscriptions must be created using a POST
Query API subscriptions have caused minor confusion as some have used the response of a GET to /subscriptions in order to connect to an existing websocket. The correct procedure it to always use a POST when connecting to a websocket, however you MAY be returned the URL of an existing websocket if the Query API deems there to be a suitable one available. The GET response should be used only for debug.

andrewbonney:
Fixes #26

### DIFF
--- a/APIs/QueryAPI.raml
+++ b/APIs/QueryAPI.raml
@@ -435,7 +435,7 @@ documentation:
           schema: ErrorSchema
   get:
     is: [paged]
-    description: List active connections to websockets providing real-time updates on resource changes
+    description: List active connections to websockets providing real-time updates on resource changes (for debug use only)
     responses:
       200:
         body:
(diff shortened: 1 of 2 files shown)


garethsb:
I went to a lot of hassle to support multiple clients connected to the same websocket subscription (potentially with different message phase too); it'd definitely have been easier not to have to!

andrewbonney:
I'd note that the reference implementation does this too and it's not intended to stop that as an option, but the intention was always for the server to handle this logic so that a client doesn't have to filter through a potentially large list of existing subscriptions to find one that matches exactly the same filters.

PR #47, opened 2018-02-27
docs: clarify trailing slash policy regarding 301 redirects

### LINKED ISSUE #15: Trailing slash policy is not clear enough
The trailing slash policy for the APIs is described here: https://github.com/AMWA-TV/nmos-discovery-registration/blob/master/docs/2.0.%20APIs.md#urls-approach-to-trailing-slashes. This describes client behaviour but fails to make important observations about server implementations as they have been tested at workshops.

A server implementing the APIs MUST support GET/HEAD (ie. read related) operations to both the trailing slash and non-trailing slash version of each resource. One of these MAY produce a 301 response causing a redirect to the other if preferred.

A server implementing the APIs MUST support PUT/PATCH/POST/DELETE/OPTIONS (ie. write related) operations on the non-trailing slash version of each resource. The implementation MAY additionally support the same methods on the trailing slash version but this is not strictly required.

The latter of these cases is particularly important as clients which receive a 301 type response when performing a PUT/POST etc will trigger a GET once redirected rather than the originally intended method. As such we have standardised on using the non-trailing slash version for all 'write' requests to avoid ambiguity between implementations.

The documentation should be enhanced to reflect the above.

andrewbonney:
Fixes #15

### DIFF
--- a/docs/2.0. APIs.md
+++ b/docs/2.0. APIs.md
@@ -85,7 +85,7 @@ In order to overcome shortcomings in some common libraries, the following requir
 
 * Clients performing requests using these methods MUST use URLs with no trailing slash present.
 * Servers implementing the APIs MUST correctly interpret requests using these methods to paths without trailing slashes present.
-* Servers implementing the APIs MAY correctly interpret requests using these methods to paths with trailing slashes present.
+* Servers implementing the APIs MAY correctly interpret requests using these methods to paths with trailing slashes present for backward compatibility.
 * Servers SHOULD NOT respond with 3xx codes for these request types.
 
 ### Error Codes & Responses
--- a/docs/2.0. APIs.md
+++ b/docs/2.0. APIs.md
@@ -71,8 +71,22 @@ A v1.1 API response may include:
 
 ### URLs: Approach to Trailing Slashes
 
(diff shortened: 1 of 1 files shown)


garethsb (changes requested):
Two comments:
1) I wonder, since OPTIONS for CORS preflight is sometimes required for GET requests (as mentioned in #43), should be described with GET and HEAD?
2) Since this change means, for PUT, POST, etc. clients now "MUST use URLs with no trailing slash", there's actually no need for the statement "Servers implementing the APIs MAY correctly interpret requests using these methods to paths with trailing slashes present" apart from backward compatibility. If that's the purpose, worth adding "For backward compatibility,"?

andrewbonney:
1. I may have to do some tests on that one. I'll post back afterwards.
2. Yes, makes sense I'll add that.

andrewbonney:
Holding off merging this until point 1 above has been properly addressed.

garethsb:
I think the conclusion in #43 was that in fact CORS pre-flight wasn't required on any of the GET requests in the NMOS APIs. If so, point 1 is probably redundant?

andrewbonney:
Yes I guess that's true. I was thinking of looking into it in case this is required for anything in the future, but it's probably fine to leave it as-is for now.

PR #56, opened 2018-04-12
raml: update to raml 1.0

### LINKED ISSUE #52: Update API specifications to use RAML 1.0
This would bring IS-04 into line with the version of RAML used by IS-05 and other work in progress specifications.

https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md

As part of this update, consider removing versions which are baked into the filenames of the example JSON files. These could be confusing where they don't reference the API version they are released as part of (e.g. example files currently include 'v1.1' but are part of the v1.2 spec).

andrewbonney:
Resolved #52

### DIFF
--- a/APIs/NodeAPI.raml
+++ b/APIs/NodeAPI.raml
@@ -1,4 +1,4 @@
-#%RAML 0.8
+#%RAML 1.0
 
 # AMWA NMOS Discovery and Registration Specification: Node API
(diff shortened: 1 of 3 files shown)


peterbrightwell (approved):
Works for me (openbrew raml2html 6.7.0)

PR #58, opened 2018-05-29
add package.json in order to help building the doc

thosil:
as we need raml2html (therefore a node.js local install), here I suggest to add a package.json and help the user to build the doc with simple npm commands

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 .DS_Store
 *.swp
 *.html
+node_modules/
+.idea/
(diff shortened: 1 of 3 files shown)


andrewbonney:
Hi. Sorry for the significant delay on this. We have a slight concern about merging this into the main spec branches given how much it's likely to need to be updated over time.

It may be possible to create a new branch specifically for 'tooling' which simply tracks the most recently released specification version (currently v1.2) and includes this, but allows us to update it independent of the specification release process. Would this work?

Alternatively, are you aware that rendered versions of the RAML are already available? The links for IS-04 are available here: https://github.com/amwa-tv/nmos/wiki/IS-04

PR #68, opened 2018-10-29
Improve existing pattern matches in schemas

### LINKED ISSUE #60: Pattern regexes in the schemas could be improved
Most of the regex begin with ``^`` and finish with ``$`` to ensure they only match the entire string, but not all. E.g. clock_internal.json has ``"pattern": "clk[0-9]+"`` where ``"pattern": "^clk[0-9]+$"`` is presumably intended. In the v1.2.x branch, the ``^``...``$`` are missing in patterns in clock_internal.json, clock_ptp.json, node.json and source_audio.json.

API versions in node.json are currently specified with ``"pattern": "v[0-9]+.[0-9]+"``. In addition to the above fix, the ``.`` is presumably supposed to match a _literal_ ``.`` so this regex should be ``"pattern": "^v[0-9]+\\.[0-9]+$"``.

Additionally in the node schema, the patterns for "chassis_id" and "port_id" use ``[-]``; I think this _is_ valid because ``-`` doesn't have its special meaning in a character-range when it's the first or last character in it, but it's unusual. Certainly in other places, like all the UUID patterns we use bare ``-``.

andrewbonney:
Resolves #60

### DIFF
--- a/APIs/schemas/clock_internal.json
+++ b/APIs/schemas/clock_internal.json
@@ -11,7 +11,7 @@
     "name": {
       "type": "string",
       "description": "Name of this refclock (unique for this set of clocks)",
-      "pattern": "clk[0-9]+"
+      "pattern": "^clk[0-9]+$"
     },
     "ref_type": {
       "description": "Type of external reference used by this clock",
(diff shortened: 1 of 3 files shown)


garethsb:
Looks good.

PR #69, opened 2018-10-29
Remove v1.1 from examples and paths in RAML

### LINKED ISSUE #59: Correct API version in examples [editorial]
Minor issue noted in comments on 8f8718fc44c88d557016fd74a993be08450cb5b7. There are two example Location response header values in RegistrationAPI.raml which probably ought to reflect the version at the top of the file, but they were last updated for v1.1. Another option would be to use "{version}" or something? 

There are other places in the v1.2 documentation that still use v1.1 endpoints as examples, e.g. [2.5. APIs - Query Parameters - Examples](https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.2.x/docs/2.5.%20APIs%20-%20Query%20Parameters.md#examples).

The example JSON responses/request files are still named for (and compatible with) v1.1 as well, though their content is also valid according to the v1.2 schemas I believe.

andrewbonney:
Resolves #59


garethsb:
That must have been painful. Looks good. :-)

PR #70, opened 2018-10-29
docs: remove requirement for manifest_href to have .sdp extension

### LINKED ISSUE #63: File extension in "manifest_href"
_Minor documentation nit..._

It would seem to be reasonable for an implementation to have the 'manifest_href' of a Sender refer to the relevant IS-05 Connection API endpoint?

However, [APIs: Server Side Implementation Notes > Node API Senders: Use with RTP](https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.2.x/docs/2.3.%20APIs%20-%20Server%20Side%20Implementation%20Notes.md#node-api-senders-use-with-rtp) says:

> When using RTP for a sender, the 'manifest_href' must be an HTTP-accessible reference to an SDP file. As-per RFC 4566, SDP files advertised via web services should use a MIME type of 'application/sdp' _and a file extension of '.sdp'. [my emphasis]_

On the other hand, the [IS-05 Connection API /transportfile endpoint](https://github.com/AMWA-TV/nmos-device-connection-management/blob/v1.0.x/APIs/ConnectionAPI.raml#L273) is understandably extension-less (to leave the door open for other transport file formats).

RFC 4566 actually seems to only have a single mention of a file extension, [Section 8.1 The "application/sdp" Media Type](https://tools.ietf.org/html/rfc4566#section-8.1), which says:

> File extension(s): The extension ".sdp" is commonly used.

Since the Media Type in a web services context is more relevant than the file extension, I suggest dropping the reference to file extension in the IS-04 documentation, or explicitly allowing the extension-less case.

andrewbonney:
Resolves #63

### DIFF
--- a/docs/2.3. APIs - Server Side Implementation Notes.md
+++ b/docs/2.3. APIs - Server Side Implementation Notes.md
@@ -25,4 +25,4 @@ Where 'href' or 'host' attributes are specified in the APIs, it is strongly reco
 
 ## Node API Senders: Use with RTP
 
-When using RTP for a sender, the 'manifest_href' must be an HTTP-accessible reference to an SDP file. As-per [RFC 4566](https://tools.ietf.org/html/rfc4566), SDP files advertised via web services should use a MIME type of 'application/sdp' and a file extension of '.sdp'.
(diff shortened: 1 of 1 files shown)


garethsb:
Thanks!

andrewbonney:
I'll merge this into v1.3-dev now, but we probably need to canvas opinion regarding whether to backport it (along with some other items in v1.3).

PR #71, opened 2018-10-29
dns-sd: add a new service type to resolve Reg API length issue

### LINKED ISSUE #34: Registration API mdns service name too long
RegistrationAPI.raml specifies an mDNS service type of "_nmos-registration".

According to RFC6763 Section 7.2, the maximum service name length for an mDNS advertisement is 16 characters when including the leading underscore, but "_nmos_registration" is 18 characters.

andrewbonney:
A potential solution to the DNS-SD service name length issue, which would take effect from v1.3 onwards. Requires further discussion.

Resolves #34

### DIFF
--- a/APIs/RegistrationAPI.raml
+++ b/APIs/RegistrationAPI.raml
@@ -15,9 +15,10 @@ documentation:
       The Registration API is exposed by NMOS discovery Nodes. It is used to publish data to the distributed registry, which can then be queried via the Query API. In smaller deployments where no distributed registry is available, the Registration API is not used and discovery responsibilities fall to individual Nodes which choose to implement Peer to Peer specification. The Registration API is a Write Only API.
   - title: DNS-SD Advertisement
     content: |
-      Registration APIs MUST produce an mDNS advertisement of the type \_nmos-registration.\_tcp. This MAY be accompanied by a unicast DNS announcement of the same type.
+      Registration APIs MUST produce an mDNS advertisement of the type \_nmos-register.\_tcp. This MAY be accompanied by a unicast DNS announcement of the same type.
+
+      Registration APIs supporting versions 'v1.2' and below MUST additionally produce an mDNS advertisement of type \_nmos-registration.\_tcp. This MAY be accompanied by a unicast DNS announcement of the same type.
       *note: RFC6763 Section 7.2 specifies that the maximum service name length for an mDNS advertisement is 16 characters when including the leading underscore, but "_nmos-registration" is 18 characters.*
-      <!-- TODO: Address the length of the mDNS advertisement type -->
 
       The IP address and port of the Registration API MUST be identified via the DNS-SD advertisement, with the full HTTP path then being resolved via the standard NMOS API path documentation.
 
--- a/docs/3.1. Discovery - Registered Operation.md
+++ b/docs/3.1. Discovery - Registered Operation.md
@@ -7,7 +7,8 @@ This document describes usage of NMOS APIs for discovery in cases where where a 
 ## Registration
 
 1. Node comes online
-2. Node scans for an active Registration API on the network using unicast and/or multicast DNS service discovery (type '\_nmos-registration.\_tcp')
+2. Node scans for an active Registration API on the network using unicast and/or multicast DNS service discovery (type '\_nmos-register.\_tcp')
+   *  Nodes which support v1.2 and earlier versions MUST additionally browse for the deprecated service type '\_nmos-registration.\_tcp'.
 3. Given multiple returned Registration APIs, the Node orders these based on their advertised priority (TXT pri), filtering out any APIs which do not support its required API version and protocol (TXT api_ver and api_proto).
 4. The Node selects a Registration API to use based on the priority, and a random selection if multiple Registration APIs with the same priority are identified (see mechanism specified in [RFC 2782](https://tools.ietf.org/html/rfc2782)).
 5. Node proceeds to register its resources with the selected Registration API.
@@ -15,6 +16,8 @@ This document describes usage of NMOS APIs for discovery in cases where where a 
 If the chosen Registration API does not respond correctly at any time, another Registration API should be selected from the discovered list. Should no further Registration APIs be available or TTLs on advertised services expired, a re-query may be performed.
 
 If no Registration APIs are advertised on a network, the Node should assume peer to peer operation unless configured otherwise. The required TXT record advertisements for this mode are described in the [Node API](../APIs/NodeAPI.raml).
+
+See the [Upgrade Path](6.0.%20Upgrade%20Path.md) for further details on supporting multiple versions simultaneously.
 
 ## Querying
 
--- a/docs/3.1. Discovery - Registered Operation.md
+++ b/docs/3.1. Discovery - Registered Operation.md
@@ -10,6 +10,7 @@ This document describes usage of NMOS APIs for discovery in cases where where a 
 2. Node scans for an active Registration API on the network using unicast and/or multicast DNS service discovery (type '\_nmos-register.\_tcp')
    *  Nodes which support v1.2 and earlier versions MUST additionally browse for the deprecated service type '\_nmos-registration.\_tcp'.
 3. Given multiple returned Registration APIs, the Node orders these based on their advertised priority (TXT pri), filtering out any APIs which do not support its required API version and protocol (TXT api_ver and api_proto).
+   *  Where a Node supports multiple API versions simultaneously, see the [Upgrade Path](6.0.%20Upgrade%20Path.md) for additional requirements in filtering the discovered API list.
 4. The Node selects a Registration API to use based on the priority, and a random selection if multiple Registration APIs with the same priority are identified (see mechanism specified in [RFC 2782](https://tools.ietf.org/html/rfc2782)).
 5. Node proceeds to register its resources with the selected Registration API.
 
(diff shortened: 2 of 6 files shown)


garethsb:
Difficult, isn't it?

This fix does lose the symmetry between Query API = _nmos-query, Registration API = _nmos-registration. And brings us closer to confusion with the nmos-parameter-registers.

I don't have a better idea though... :-/

garethsb:
Data point: The Bonjour/mDNSResponder implementation (I confirmed with a couple of versions on Windows and Linux) seems to handle a too-long name in mDNS mode, but rejects it point blank in unicast DNS-SD mode.

Demo...

```
# dns-sd -B _1234567890123456._tcp
Browsing for _1234567890123456._tcp
...
^C
# dns-sd -B _123456789012345._tcp example.com
Browsing for _123456789012345._tcp.example.com
...
^C
# dns-sd -B _1234567890123456._tcp example.com
Browsing for _1234567890123456._tcp.example.com
DNSService call failed -65540
```

-65540 is ``kDNSServiceErr_BadParam``

FWIW, Avahi (even via the Bonjour/mDNSResponder compatibility layer) seems to handle the longer name in both modes.

----

For reference, see a comment in the source code of [mDNSResponder-878.30.4/mDNSShared/uds_daemon.c](https://opensource.apple.com/source/mDNSResponder/mDNSResponder-878.30.4/mDNSShared/uds_daemon.c.auto.html) around line 2568.
```C
    // For over-long service types, we only allow domain "local"
    if (temp.c[0] > 15 && domain[0] == 0) mDNSPlatformStrLCopy(domain, "local.", sizeof(domain));
```
Or more to the point, the code in [mDNSResponder-878.30.4/mDNSCore/DNSCommon.c](https://opensource.apple.com/source/mDNSResponder/mDNSResponder-878.30.4/mDNSCore/DNSCommon.c.auto.html) around line 1042.
```C
    src = type->c;                                      // Put the service type into the domain name
    len = *src;
    if (len < 2 || len > 16)
    {
        LogMsg("Bad service type in %#s.%##s%##s Application protocol name must be underscore plus 1-15 characters. "
               "See <http://www.dns-sd.org/ServiceTypes.html>", name->c, type->c, domain->c);
    }
    if (len < 2 || len >= 0x40 || (len > 16 && !SameDomainName(domain, &localdomain))) return(mDNSNULL);
    if (src[1] != '_') { errormsg = "Application protocol name must begin with underscore"; goto fail; }
```

andrewbonney:
Likely makes it a bit more critical to make this available from v1.3 rather than a later version given what's being proposed in https://github.com/AMWA-TV/nmos-discovery-registration/pull/75

garethsb:
Yep.

garethsb:
So if a Registry were to support both v1.2 and v1.3, then it must have *both* advertisements?

Should those advertisements both include both versions?

_nmos-register._tcp, with "api_ver=v1.3" or "api_ver=v1.2,v1.3"?
_nmos-registration._tcp, with "api_ver=v1.2" or "api_ver=v1.2,v1.3"?

Does a Node that supports both v1.2 and v1.3, need to get *both* _nmos-register._tcp and _nmos-registration advertisements in order to choose one with an appropriate "pri"?
(I'm assuming that higher (lower numerically) "pri" on a lower API version is 'better' than a lower "pri" on a higher API version?)

(One more question...) Is there any downside to *recommending* that even Registries that don't support v1.3, should also advertise with _nmos-register._tcp?

andrewbonney:
Yes, if a registry supports v1.2 and v1.3 it must have both advertisements in order to maintain backwards compatibility. I'd suggest that the nmos-register type could have all supported versions listed, but nmos-registration could be limited to listing v1.2 and below as all v1.3 compatible Nodes must be able to identify the new service type.

If a Node isn't locked to a specific API version I'd have suggested it would always be best using the highest API version available, then using 'pri' only within the API version it is interested in.

I see no downside to recommending using nmos-register in earlier versions, we just can't rely on existing Node implementations (v1.2 and below) to pick up on it.

I'll add notes on these before this branch gets merged, and of course there's plenty here as candidates for addition to the test suite.

garethsb:
> If a Node isn't locked to a specific API version I'd have suggested it would always be best using the highest API version available, then using 'pri' only within the API version it is interested in.

This is interesting... I can imagine situation that a Registry might want to declare v1.2 with a high priority, but v1.3 with a lower priority. Is that possible somehow?

andrewbonney:
It certainly wasn't the original intent as you'd need to make two separate DNS-SD advertisements for the same registry, with each one listing just the relevant version(s). I'd suggest if there are use cases for it then the behaviour would need to be documented as a new feature.

garethsb:
If version is more important than "pri", that simplifies the life of a v1.3 Node because if it discovers *any* v1.3 Registration API of an acceptable"pri" via _nmos-register._tcp, it doesn't need to go on and browse for _nmos-registration._tcp.

andrewbonney:
@garethsb-sony could you see if the two commits I've added resolve your queries? I've kept the latter commit separate as I think it's probably worth backporting that onto earlier versions.

garethsb:
Thanks, those new sentences reflect what you've said here.

Considering version ahead of priority will be a change for our implementations; doable but a change. And I have a follow-on question... if the Node discovers a v1.3 Registration API with pri=42, and a v1.2 Registration API with pri=37, according to this resolution it should pick the v1.3 API. If it later (or immediately) fails to connect to that API, should it switch to the v1.2 API, or should it rescan? I.e. should the priority list of Registration APIs that the Node maintains be of only the highest version it finds, or all versions?

> I see no downside to recommending using nmos-register in earlier versions, we just can't rely on existing Node implementations (v1.2 and below) to pick up on it.

I'm not sure if we want to add this also? I'm only thinking that if Registry vendors did this, it'd be a quick fix for v1.2 Nodes to support unicast DNS-SD via Bonjour/mdnsResponder (and any other DNS-SD implementation that follows the RFC). But there are other ways round this...

andrewbonney:
> Considering version ahead of priority will be a change for our implementations; doable but a change. And I have a follow-on question... if the Node discovers a v1.3 Registration API with pri=42, and a v1.2 Registration API with pri=37, according to this resolution it should pick the v1.3 API. If it later (or immediately) fails to connect to that API, should it switch to the v1.2 API, or should it rescan? I.e. should the priority list of Registration APIs that the Node maintains be of only the highest version it finds, or all versions?

In the absence of another suitable v1.3 API and without a re-scan, I'd suggest switching to use the v1.2 API is fine (subject to DELETEing things first if the first heartbeat/registration with the v1.2 API returns a 200 code, given the version conflict). If the Node is operating with a version-locked configuration to ensure it only ever operates at v1.3 then obviously this wouldn't apply.

Whether a 're-scan' should or shouldn't be performed is probably a more general topic for discussion. I suspect at present a mix of approaches is used to that.


> I'm not sure if we want to add this also? I'm only thinking that if Registry vendors did this, it'd be a quick fix for v1.2 Nodes to support unicast DNS-SD via Bonjour/mdnsResponder (and any other DNS-SD implementation that follows the RFC). But there are other ways round this...

I missed the note about registries which only implement up to v1.2 when dealing with this. I wonder if that requires a more forward looking patch to the v1.0 to v1.2 spec docs as opposed to including it with anything that goes directly into v1.3 (ie. a separate PR). That way we don't need to specify anything new in the older versions, but can provide a pointer to the new spec around how to implement the new service type.

docs/6.0. Upgrade Path.md
```diff
@@ -8,10 +8,12 @@ API versioning is specified in the [APIs](2.0.%20APIs.md) documentation, with pr
 
 ## Requirements for Nodes (Node APIs)
 
-Implementers of the Node API must support at least one API version, and may support more than one at a time. Note however that a Node must only perform interactions with a Registration API at a single version. Nodes implementing multiple API versions may provide a user-configurable choice for which API version to register and/or query using.
+Implementers of the Node API must support at least one API version, and may support more than one at a time. Note however that a Node must only perform interactions with a Registration API at a single version. Nodes implementing multiple API versions may provide a user-configurable choice for which API version to register and/or query using. Otherwise it is expected that the Node will select a Registration API which supports the highest API version which the Node also supports. The API version should be selected before considering the Registration API's configured \'priority\'.
```
btrost-harmonicinc: Do you want to use "MUST"/"MAY" in this section rather than "must"/"may"?

andrewbonney: Yes, I should probably fix that whilst editing. There are likely other cases of this in the docs which need resolving too.

docs/3.1. Discovery - Registered Operation.md
```diff
@@ -10,20 +10,20 @@ This document describes usage of NMOS APIs for discovery in cases where where a
 2. Node scans for an active Registration API on the network using unicast and/or multicast DNS service discovery (type '\_nmos-register.\_tcp')
    *  Nodes which support v1.2 and earlier versions MUST additionally browse for the deprecated service type '\_nmos-registration.\_tcp'.
 3. Given multiple returned Registration APIs, the Node orders these based on their advertised priority (TXT pri), filtering out any APIs which do not support its required API version and protocol (TXT api_ver and api_proto).
+   *  Where a Node supports multiple API versions simultaneously, see the [Upgrade Path](6.0.%20Upgrade%20Path.md) for additional requirements in filtering the discovered API list.
```
garethsb, line 13: @andrewbonney I think the next bullet (4) could be usefully clarified as well, e.g. inserting "of the same API version" after "multiple Registration APIs". I'm also not sure it's been helpful mentioning RFC 2782 here since af9b50d56974a442c3ba4a95511ff3529e252752 (v1.1).

(The same would then apply to the **Querying** section below.)

APIs/RegistrationAPI.raml
```diff
@@ -34,6 +34,8 @@ documentation:
 
       The DNS-SD advertisement MUST be accompanied by a TXT record of name 'api\_ver'. The value of this TXT record is a comma separated list of API versions supported by the server. For example: 'v1.0,v1.1,v2.0'. There should be no whitespace between commas, and versions should be listed in ascending order.
 
+      When both defined service types are advertised, the following practice MUST be observed. All supported API versions MUST be listed against the service type \_nmos-register.\_tcp. All supported API versions up to and including (but not beyond) v1.2 MUST be listed against the service type \_nmos-registration.\_tcp.
```
garethsb: Thinking about it, I see no need to limit the versions advertised via _nmos-registration._tcp, so for simplicity (the values for both record names can be identical) I'd prefer to delete "(but not beyond)" (and replace with "at least" if you like).

andrewbonney:
@garethsb-sony does this resolve all of the open queries around the new service type?

garethsb:
I believe so!!! 🤞
Thank you, Andrew.

PR #75, opened 2018-12-14
Make peer to peer Node mDNS optional, and unicast DNS preferred

andrewbonney:


peterbrightwell:
@andrewbonney: can we merge this now?

andrewbonney:
Not quite. Given the issues in some common DNS-SD libs we think it'd be worth getting https://github.com/AMWA-TV/nmos-discovery-registration/pull/71 in at the same time. I just need to add a couple of clarifications to that which I'll attempt to get done today.

garethsb:
Should there also be a few changes to [3.0 Discovery](https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.2.x/docs/3.0.%20Discovery.md) advice? E.g. clarifying operation in mDNS or unicast DNS-SD only modes, and the use of 'api_ver' as well as 'pri' to prioritise the results.

andrewbonney:
@garethsb-sony were your queries here addressed by https://github.com/AMWA-TV/nmos-discovery-registration/pull/71 or do you think something further is required for this PR?

garethsb:
I think everything addressed between #71 and your commits in this PR on 14 Dec. Thanks, Andrew!

PR #77, opened 2019-01-16
registration: add informative note on fix for over-sized service type

andrewbonney:
Adds a note to the documentation of release spec versions to indicate how issues with the over-sized _nmos-registration._tcp name can be resolved.

Do not merge this until v1.3 of IS-04 (which includes the resolution) has been released.


andrewbonney:
Merging this now as v1.3 should be released alongside v1.2.2

PR #78, opened 2019-01-16
docs: clarify definition of a URN subclassification

andrewbonney:
Requires backporting to v1.2 and below

### DIFF
--- a/docs/2.1. APIs - Common Keys.md
+++ b/docs/2.1. APIs - Common Keys.md
@@ -82,12 +82,21 @@ See [Use of URNs](#use-of-urns) below for additional requirements regarding thei
 
 All Uniform Resource Names (URNs) specified within the scope of the NMOS specifications MUST use the namespace 'urn:x-nmos:'. Manufacturers MAY use their own namespaces to indicate values of URN which are not currently defined within the NMOS namespace.
 
-Where a URN begins 'urn:x-nmos:' it MAY include one or more 'subclassifications'. Subclassifications are defined as the portion of the URN which follows the first occurrence of a '.'. By convention, URNs which begin 'urn:x-nmos:' will never include a further ':' after the first occurrence of a '.'.
+Where a URN begins 'urn:x-nmos:' it MUST use the following construction:
+*   The 'root' of the URN is entirely delimited by ':' characters
+*   The URN MAY include one or more 'subclassifications'
+*   The URN MAY include a version number
 
-For example, the URN 'urn:x-nmos:transport:rtp' above has two subclassifications of 'ucast' and 'mcast'.
+Subclassifications are defined as the portion of the URN which follows the first occurrence of a '.', but prior to any '/' character. By convention, URNs which begin 'urn:x-nmos:' will never include a further ':' after the first occurrence of a '.'.
+
+For example, the URN root 'urn:x-nmos:transport:rtp' above has two subclassifications of 'ucast' and 'mcast'.
 
 API clients interpreting URNs MUST correctly recognise any as-yet undefined subclassifications as being associated with the URN root which appears prior to the first '.' character.
 
 For example, if filtering by the format 'urn:x-nmos:format:video', the filter is required to include any as-yet undefined subclassifications such as 'urn:x-nmos:format:video.raw'.
+
+Versions are defined as the portion of the URN which follows the first occurrence of a '/', up to and including the remainder of the URN. Any '.' characters after this point indicate a version delimiter.
+
+For example, the URN 'urn:x-nmos:control:sr-ctrl/v1.0' can be separated into a root 'urn:x-nmos:control:sr-ctrl' and version 'v1.0'.
 
 Query API clients MUST be tolerant to URNs which have not yet been defined, but which may be added in later API versions.
--- a/docs/2.1. APIs - Common Keys.md
+++ b/docs/2.1. APIs - Common Keys.md
@@ -83,7 +83,7 @@ See [Use of URNs](#use-of-urns) below for additional requirements regarding thei
 All Uniform Resource Names (URNs) specified within the scope of the NMOS specifications MUST use the namespace 'urn:x-nmos:'. Manufacturers MAY use their own namespaces to indicate values of URN which are not currently defined within the NMOS namespace.
 
 Where a URN begins 'urn:x-nmos:' it MUST use the following construction:
-*   The 'root' of the URN is entirely delimited by ':' characters
+*   The URN 'base' is a sequence of one or more names delimited by ':' characters, including the namespace
 *   The URN MAY include one or more 'subclassifications'
 *   The URN MAY include a version number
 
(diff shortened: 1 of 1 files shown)


garethsb:
"First '.'", or "first '.' after the last ':'"?

Simple case is fine either way. These two share a common 'top-level category' either way:

* urn:x-nmos:foo:bar.baz
* urn:x-nmos:foo:bar.qux

But do these two share a common 'top-level category' or not?

* urn:x-nmos:foo.bar:baz
* urn:x-nmos:foo.qux:quux

andrewbonney:
I'd suggest the former. I think if we end up with something that looks like the latter it's going to be confusing for a whole range of reasons, so we should try to avoid doing it anyway. If at any point such a classifier was introduced we could consider use of another term for the entity which follows the final dot.

garethsb:
i completely agree that finding a '.' before the final ':' is confusing.

So, either we ban that, or we define the meaning of top-level category and subclassification clearly.

Do you propose that the URN "urn:x-nmos:foo.bar:baz" has a top-level category of "urn:x-nmos:foo", and a subclassification of "bar:baz"?

andrewbonney:
Yes. If that isn't coming across clearly enough via this modification, would you mind suggesting some alternative text?

garethsb:
No, that's how I read it, which is why I objected to it. In my mind, ':' has been a 'stronger' separator than '.'.

peterbrightwell:
I agree - don't use a ':' after you've used a '.'.   So that example "urn:x-nmos:foo.bar:baz" shouldn't be used.

garethsb:
> shouldn't

Andrew will know what I'm going to say...
We have to say MUST NOT, or accept that SHOULD NOT means MAY. :-)

peterbrightwell:
Agreed but that was just an informative example :).

garethsb:
So where does that leave us with this PR?

andrewbonney:
@garethsb-sony could you see if that is any clearer? Obviously this is applicable beyond IS-04, so it's one of many things which is prime for refactoring.

docs/2.1. APIs - Common Keys.md
```diff
@@ -80,6 +76,18 @@ Note: From v1.3 onwards, permitted device types are not versioned and are instea
...
-As with formats (described above), subclassifications of these URNs must be recognised correctly as members of these top level categories.
+See [Use of URNs](#use-of-urns) below for additional requirements regarding their handling.
+
+## Use of URNs
+
+All Uniform Resource Names (URNs) specified within the scope of the NMOS specifications MUST use the namespace 'urn:x-nmos:'. Manufacturers MAY use their own namespaces to indicate values of URN which are not currently defined within the NMOS namespace.
+
+Where a URN begins 'urn:x-nmos:' it MAY include one or more 'subclassifications'. Subclassifications are defined as the portion of the URN which follows the first occurrence of a '.'. By convention, URNs which begin 'urn:x-nmos:' will never include a further ':' after the first occurrence of a '.'.
+
+For example, the URN 'urn:x-nmos:format:rtp' above has two subclassifications of 'ucast' and 'mcast'.
```
garethsb: Little typo to fix: format --> transport

garethsb:
Looks good to me! Thanks, Andrew.

docs/2.1. APIs - Common Keys.md
andrewbonney: Ta. A little spec blindness there despite re-reading

garethsb:
I think the final piece here is deciding the answer to https://github.com/AMWA-TV/nmos-device-connection-management/issues/57 i.e. whether IS-04 and IS-05 transport type values should be identical or not.

If not, IS-05 v1.1 needs words to indicate that a subclassification should be removed from the IS-04 ``"transport"`` value to make the ``/transport_type`` value.

andrewbonney:
Yes, I'll look to merge this into the dev branch tomorrow and we can continue that one via the referenced issue.

garethsb:
I've just thought of something...

The NMOS specs use *versioned URNs* in several places, e.g. ``urn:x-nmos:control:sr-ctrl/v1.0``. Several are listed in the [NMOS Parameter Registers](https://github.com/AMWA-TV/nmos-parameter-registers).

Do we view ``urn:x-nmos:control:sr-ctrl/v1.0`` as a subclassification of ``urn:x-nmos:control:sr-ctrl/v1``?
I suppose it makes a kind of sense... 😬

andrewbonney:
Good point. I think I might have one more go at it, splitting the potential URN into a root, subclassification and version, with the last two components being optional.

andrewbonney:
That should cover the versioning now. I'm just going to have to be a little careful when backporting this to older docs as it effectively introduces some normative requirements which weren't present previously under certain conditions.

docs/2.1. APIs - Common Keys.md
```diff
@@ -82,12 +82,21 @@ See [Use of URNs](#use-of-urns) below for additional requirements regarding thei
 
 All Uniform Resource Names (URNs) specified within the scope of the NMOS specifications MUST use the namespace 'urn:x-nmos:'. Manufacturers MAY use their own namespaces to indicate values of URN which are not currently defined within the NMOS namespace.
 
-Where a URN begins 'urn:x-nmos:' it MAY include one or more 'subclassifications'. Subclassifications are defined as the portion of the URN which follows the first occurrence of a '.'. By convention, URNs which begin 'urn:x-nmos:' will never include a further ':' after the first occurrence of a '.'.
+Where a URN begins 'urn:x-nmos:' it MUST use the following construction:
+*   The 'root' of the URN is entirely delimited by ':' characters
```
garethsb: I think 'root' needs explanation as its a very generic term, or maybe 'base' or 'base classification' (cf. "base url") is clearer?
I understand what you're trying to say with "entirely delimited by" but I wonder if it could also be expressed more clearly.

How about:

* The 'base' URN continues with a sequence of one or more names delimited by ':' characters.

docs/2.1. APIs - Common Keys.md
```diff
@@ -82,12 +82,21 @@ See [Use of URNs](#use-of-urns) below for additional requirements regarding thei
...
+Subclassifications are defined as the portion of the URN which follows the first occurrence of a '.', but prior to any '/' character. By convention, URNs which begin 'urn:x-nmos:' will never include a further ':' after the first occurrence of a '.'.
+
+For example, the URN root 'urn:x-nmos:transport:rtp' above has two subclassifications of 'ucast' and 'mcast'.
 
 API clients interpreting URNs MUST correctly recognise any as-yet undefined subclassifications as being associated with the URN root which appears prior to the first '.' character.
 
 For example, if filtering by the format 'urn:x-nmos:format:video', the filter is required to include any as-yet undefined subclassifications such as 'urn:x-nmos:format:video.raw'.
 
+Versions are defined as the portion of the URN which follows the first occurrence of a '/', up to and including the remainder of the URN. Any '.' characters after this point indicate a version delimiter.
+
```
garethsb, line 99: Maybe add something like:

By convention, for URNs defined by NMOS specifications, the version portion is formatted as 'v<#MAJOR>.<#MINOR>' and should be handled as described in [NMOS API Versioning](https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.3-dev/docs/2.0.%20APIs.md).

docs/2.1. APIs - Common Keys.md
garethsb: I almost want to be done with it and add a regex or ABNF grammar here, but I know that's excessive. ;-)

garethsb (commented):
Nearly there...

docs/2.1. APIs - Common Keys.md
andrewbonney: I came very close... Unfortunately as they're not universally understood I think it'd still need the text.

andrewbonney:
I'll squash these all down into one commit once they're ready to merge.

docs/2.1. APIs - Common Keys.md
```diff
@@ -80,6 +76,27 @@ Note: From v1.3 onwards, permitted device types are not versioned and are instea
...
+Where a URN begins 'urn:x-nmos:' it MUST use the following construction:
+*   The URN 'base' is a sequence of one or more names delimited by ':' characters, including the namespace
+*   The URN MAY include one or more 'subclassifications'
+*   The URN MAY include a version number
+
+Subclassifications are defined as the portion of the URN which follows the first occurrence of a '.', but prior to any '/' character. By convention, URNs which begin 'urn:x-nmos:' will never include a further ':' after the first occurrence of a '.'.
+
+For example, the URN root 'urn:x-nmos:transport:rtp' above has two subclassifications of 'ucast' and 'mcast'.
+
+API clients interpreting URNs MUST correctly recognise any as-yet undefined subclassifications as being associated with the URN root which appears prior to the first '.' character.
```
garethsb: Suggest change root to 'base'

garethsb (approved):
LGTM

PR #81, opened 2019-01-29
query: add explicit 501 requirements for all SHOULD/MAY features

andrewbonney:
Includes a mechanism to reduce the overhead of Receiver subscription implementation from v1.3 onwards.

### DIFF
--- a/APIs/NodeAPI.raml
+++ b/APIs/NodeAPI.raml
@@ -66,7 +66,9 @@ documentation:
 
   - title: Receiver Subscriptions
     content: |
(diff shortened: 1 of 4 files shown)


garethsb:
I haven't looked through for other places where 501 could be described explicitly, but these changes all look good to me.

Do you think the test suite could report a 501 on e.g. IS-04 v1.2 Node API /target as a ``WARNING`` rather than a ``FAIL``? Cf. https://github.com/AMWA-TV/nmos-testing/issues/114

andrewbonney:
That sounds fair. Whilst (to the letter of the spec) Nodes need to implement it up to and including v1.2, we know that many don't (and likely won't need to) given what IS-05 now provides.

Issue #67, opened 2018-10-09
flow_id in sender schema v1.0.x vs v1.1.x/v1.2.x


jasperpeeters:
In **v1.1.x** and **v1.2.x** the ```flow_id``` of a sender can be ```null``` regarding the ```sender.json```-schema. However, in **v1.0.x** only a valid uuid is allowed.

How to deal with this regarding the output of the Query API?

E.g. registering a **v1.2** sender with ```flow_id: null```, what should be the QueryAPI output if I ```GET``` ```<...>/query/v1.0/senders```? 

Should this resource be ignored for the output or should the resource be displayed even if the output of the API is not perfectly valid regarding the current active schema? 

Both alternatives look a bit inconsistent in their own way..

garethsb:
Here's the relevant bits of a previous e-mail exchange I had with @andrewbonney.

At 20 June 2017 17:09, @garethsb-sony wrote:
> How about the change that allows a Sender’s “flow_id” to be null in v1.1? When downgrading a v1.1 Sender with “flow_id”: null, what value should be used in a v1.0 representation of the Sender?

At 20 June 2017 17:14, @andrewbonney wrote:
> The Sender flow_id one is unfortunate and I hadn’t clocked it as an issue until now. I don’t have a good solution to that off the top of my head, and our implementation (whilst I don’t think it ever sets that to null) would certainly produce an output that isn’t technically v1.0 valid at present.

Our implementation (sony/nmos-cpp) does the same; the sender will have ``"flow_id": null``.

andrewbonney:
I should add, I'd be happy for the guidance to be omitting the offending Senders if preferred. Either way we should add this case to the docs - I'd clearly failed to log it as an issue on here previously.

I doubt there will be a massive number of v1.0 Query API clients developed at this point, but we should still aim to have a consistent approach.

andrewbonney:
I believe this was resolved by https://github.com/AMWA-TV/nmos-discovery-registration/pull/83 and will be resolved in the v1.2.x branch by https://github.com/AMWA-TV/nmos-discovery-registration/pull/91

PR #83, opened 2019-01-31
docs: add explicit documentation of inter-version key removals

andrewbonney:


garethsb:
Your lists are the same as mine in [sony/nmos-cpp api_downgrade](https://github.com/sony/nmos-cpp/blob/master/Development/nmos/api_downgrade.cpp#L67-L123) (except that I don't bother removing "active" from Receiver "subscription", because the schema doesn't have ``"additionalProperties": false`` so still validates).

andrewbonney:
Ok. I based this off the current state of the test suite mechanism, but I suspect that one could go either way.

andrewbonney:
Added a section on keys which may have unexpected values as a result of version translations. This relates to https://github.com/AMWA-TV/nmos-testing/issues/114 and https://github.com/AMWA-TV/nmos-discovery-registration/issues/67

Whilst the potential for URN enums to expand in the future was already documented (for example given they could include manufacturer defined types), this may be an indicator that a better mechanism for the Query API in a future version may be to expose only matching versioned resources at each API version, with the explicit downgrade query parameter providing the means to bridge across API versions at the client's discretion.

PR #84, opened 2019-01-31
Refactor content of RAML file headers into docs directory

andrewbonney:
I'm keen to refactor the normative text at the top of each RAML document into the docs/ directory in order to reduce the fragmentation in the specification (and as such the possibility for errors in updating it or creating implementations). So far I've simply copied text over into appropriate locations and performed some de-duplication. This is a starting point, but subject to it being considered a useful thing to do, I'll proceed to include the following additional steps within this PR:

* Refactoring normative text which exists primarily in JSON schemas into the docs/ directory (for example around Sender/Receiver subscription behaviour)
* Updating the docs/ directory to make better use of normative language as-per https://github.com/AMWA-TV/nmos-audio-channel-mapping/blob/v1.0-dev/docs/1.0.%20Overview.md#use-of-normative-language

Whilst the intention is to make no normative changes as part of this process, as the changes will hit so many aspects of the specification I don't intend to backport this refactoring to earlier specification versions. It will also be more difficult to backport changes in the future (below v1.3) as a result.

Feedback appreciated.

PR #85, opened 2019-02-04
registration: add explicit 409 code for version conflicts

andrewbonney:
This is the start of an addition to aid Nodes which support multiple API versions at once. A Node must only interact with one API version at a time, and must remove itself from a registry entirely if it intends to change version. At present a Registration API can only respond with a generic '400' in this case, which doesn't assist the Node in recovering from a recoverable error.


APIs/RegistrationAPI.raml
```diff
@@ -149,6 +153,10 @@ documentation:
         description: Returned when the requested Node does not exist or has been garbage collected
         body:
           type: ErrorSchema
+      409:
```
garethsb, line 146: Should this also apply to the (debug) get request?

garethsb (commented):
Looks good, and explains the rationale we've discussed clearly.

Would this preclude other kinds of 'conflict' being indicated by a 409?

I [currently respond with a 409](https://github.com/sony/nmos-cpp/blob/fb0d6c9ac7cc640847ad17851072136fadf88f21/Development/nmos/registration_api.cpp#L406-L444) for an additional 5 kinds of conflict:

* request has same id as an existing registration, but different type
* request contains different parent resource id than existing registration (moving a source from one device to another seems like a bad idea to me)
* request has older version timestamp than existing registration
* request has parent resource id which is in the registry but of the wrong type
* request has parent resource id which is in the registry but of the wrong API version

It seems to me we have three classes of failed requests:

* JSON syntax is in error, or JSON schema validation failed, which truly indicate a *Bad Request*
* API version *Conflict*, which can likely be resolved automatically by the Node with a DELETE (or wait for GC) and a retry POST
* Ones that fall on a spectrum between these which indicate a system issue of some kind (e.g. inconsistent clocks, duplicate UUIDs, other fail-over ephemera, ...)

APIs/RegistrationAPI.raml
andrewbonney: I was unsure on that one. It's certainly not well defined at present. 409 is definitely better than a 404. The alternative would be to mirror the Query API behaviour, but that's perhaps unnecessary.

andrewbonney:
So far I've captured those other cases under 400 (except 'request has older version timestamp than existing registration' which I could add). The reason for using that code rather than 409 was to ensure that 409 was reserved for classes of error that the Node was likely to be able to resolve via the DELETE/POST mechanism. As you describe the others fall into a greyer area which likely require interaction to resolve, and in many cases might be better caught by the test suite.

The greyer issues could always be handled via another error code, but if they're all things which would require a human to investigate then I'd have thought we might be best 'keeping it simple' in terms of response codes, but ensure that the JSON error responses have sufficient information for diagnosis.

garethsb:
> require a human to investigate then I'd have thought we might be best 'keeping it simple' in terms of response codes, but ensure that the JSON error responses have sufficient information for diagnosis.

That makes sense, and is nicely captured by https://github.com/AMWA-TV/nmos-discovery-registration/pull/85/commits/793eda6c0100e38e138db5d57e1a87383a37e7cb. It'd be good to have a pointer to that from the RAML, which currently sounds a bit black and white:
https://github.com/AMWA-TV/nmos-discovery-registration/blob/7df84c609e47ca09b97a04499fbad79591c79053/APIs/RegistrationAPI.raml#L78-L79

It'd be nice to have some other implementors chime in on this one.

andrewbonney:
I'm personally keen to follow the lead of IS-08 and move to having as little additional descriptive detail in the RAML as possible (see the starting point in https://github.com/AMWA-TV/nmos-discovery-registration/pull/84), however if that's accepted as a general approach I'd expect we could link from the RAML to a more fully featured description in the Markdown.

APIs/RegistrationAPI.raml
andrewbonney: I've added these for the other cases now, including the DELETE which I'd missed first time around

garethsb:
I like this, I'll update our implementation to follow. I think it just needs a one-sentence summary in the RAML that indicates syntax and schema validation aren't the only possible reasons for a 400.

garethsb:
After looking at implementation I have some queries:
* 409 Conflict for the GET request reads a bit oddly now I think about it... (though it is consistent)
* We should specify what the Location response header should be
  * For the POST requests, I currently return the ``/x-nmos/registration/{version}/resource/{resourceType}/{resourceId}`` URL, which makes sense, right? (Version is the actual resource version, obviously)
  * For the DELETE and GET request (if we stick with 409 for that), the same
  * For the heartbeat, should it be the ``/x-nmos/registration/{version}/health/nodes/{nodeId}`` URL?

andrewbonney:
Ok. I still think the GET behaviour needs clarifying as to my eyes that's only properly defined on the Query API side, most likely given the Reg API side is only for debug and so hasn't had much attention. This would either have to return a 404, perform a translation as in the Query API, or just display the same objects no matter which version they were requested via. The 'best' text I've identified so far suggests a similar behaviour to the Query API, which I'd assume is what most implementations have done.

Yes, the POST Location header makes sense, and matches what's in the spec for a successful request. The DELETE and heartbeat seem to make sense too. As for the GET that may depend upon the result of the above.

APIs/RegistrationAPI.raml
```diff
@@ -82,13 +82,16 @@ documentation:
...
         body:
           type: ErrorSchema
       409:
-        description: Returned when the resource already exists in the registry at a different API version
+        description: Returned when the resource already exists in the registry at a different API version. The correct API version is identified via the Location header.
         body:
           type: ErrorSchema
+        headers:
+          Location:
+            example: /x-nmos/registration/{version}/resource/nodes/3b8be755-08ff-452b-b217-c9151eb21193/
```
garethsb, line 67: From [URLs: Approach to Trailing Slashes](https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.3-dev/docs/2.0.%20APIs.md#urls-approach-to-trailing-slashes):
> For consistency and in order to adhere to how these APIs are specified in RAML, the 'primary' path for every resource has the trailing slash omitted.

I realise that applies to the other pre-existing Location examples too.

andrewbonney: Yes... Unfortunately the spec as a whole has a habit of presenting URLs with a slash in examples, but requiring that actual interactions with them remove the slash. I was conflicted as to whether to 'correct' this in these new cases or not.

garethsb:
> Ok. I still think the GET behaviour needs clarifying as to my eyes that's only properly defined on the Query API side, most likely given the Reg API side is only for debug and so hasn't had much attention. 

Agreed.

> This would either have to return a 404, 

This was my current implementation based on our discussion last July.

> perform a translation as in the Query API,

IMHO, no. Or we might as well just merge Registration and Query API.

> or just display the same objects no matter which version they were requested via.

IMHO, also no, since they wouldn't validate against schema.

> The 'best' text I've identified so far suggests a similar behaviour to the Query API, which I'd assume is what most implementations have done.

(I don't understand what you mean by "'best' text"?)

IMHO, the choice should be between 400 and 409. I'm actually coming back round to the idea that the two debug GET requests (resource and health) should return 409 Conflict for consistency with the POST and DELETE endpoints. While the POST and DELETE requests will only be made by the Node, the GET requests are presumbly also only made by a user that 'knows what they're doing', and providing the Location header in the response is as helpful here, as in those POST/DELETE requests.

APIs/RegistrationAPI.raml
garethsb: OK ☹️
Add a low-priority issue to fix all such examples?

andrewbonney: Added to https://github.com/AMWA-TV/nmos-discovery-registration/issues/87

andrewbonney:
The text I was referring to was from the first line of https://amwa-tv.github.io/nmos-discovery-registration/tags/v1.2.1/html-APIs/RegistrationAPI.html#versioning. As you note however it isn't always possible to meet the schemas of earlier versions.

Perhaps this needs a separate clarification for behaviour in v1.2 and below regarding preference to use a 404 for debug responses, but for v1.3+ to switch to a 409 as it can provide the additional debug data afforded by the Location header.

garethsb:
I think 409 Conflict should also apply to Query API for GET and DELETE /subscription/{subscriptionId} when request API version is not equal to the subscription's API version, and also to GET /{resourceType}/{resourceId} where the request API version and query.downgrade version (if present) are higher than the resource's API version.

andrewbonney:
That makes sense.

PR #100, opened 2019-06-07
Proposal for IS-07 inclusion in IS-04

### LINKED ISSUE #88: Add signalling for IS-07 data via Source attributes
As described in https://github.com/AMWA-TV/nmos-event-tally/blob/v1.0.x/docs/4.0.%20Core%20models.md#2-is-04-highlights IS-07 requires an additional Source attribute. In order to add this we need to ensure that the 'urn:x-nmos:format:data' URN is accompanied by a defined way to differentiate between different types of 'data' (including but not limited to IS-07 and SDI ancillary). Consideration should be given to this mechanism's extensibility in order to avoid unnecessary future version increments.

garethsb:
This PR is just @andrewbonney's https://github.com/bbc/nmos-discovery-registration/pull/7 retargeted at the base repo. See discussion in that PR leading to the current proposal. It would resolve #88.

This proposal is reflected into https://github.com/AMWA-TV/nmos-event-tally/pull/55 on IS-07 v1.0.x.


APIs/schemas/source_data.json
```diff
@@ -0,0 +1,29 @@
...
+      ],
+      "properties": {
+        "format": {
+          "description": "Format of the data coming from the Source as a URN",
+          "type": "string",
+          "enum": [
+            "urn:x-nmos:format:data"
+          ],
+          "format": "uri"
+        },
```
mjeras, line 21: One thing I like about IS-04 classes (unlike IS-05 transport parameters :( ) is that the type of the polymorphic objects can be deduced from a single field, e.g. "format": "urn:x-nmos:format:audio" would indicate that it is an audio flow or source. That is extremely useful when deserializing in typed objects in a generic way. 
Unfortunately, this breaks that concept as there is an overlap in what "format": "urn:x-nmos:format:data" designates (it could be either source_generic.json or source_data.json). 
There would be a workaround for IS-07 sources as they would have "event_type" and one could look for its existence, but it doesn't cover the most general case. (I assume "event_type" is not required to make this extensible to other specs that might use "urn:x-nmos:format:data"). But this relates to my other comment on this PR.

mjeras (commented):
I think there is a difference in approach ho the fields are added to IS-04 in this PR and how the transport parameters were extended for IS-07 in IS-05. In that case we introduced a set of generic WS/MQTT fields and then IS-07 added its extensions. That also avoided direct references of IS-07 from IS-05.
However, this is not a case here - in this PR we are adding optional fields that are defined by referencing IS-07.
At this point I am not sure which one is the right approach (or is there a third one, which would be better) - I am merely stating the discrepancy...

andrewbonney:
I agree it would be nice to avoid the coupling between IS-04 and 07. In the long run this sort of additive change could be documented purely in the parameter registers, potentially avoiding a version bump if we're careful about not disrupting existing use cases. I suspect it may be too much work to get that approach sorted in time for this release however.

APIs/schemas/source_data.json
andrewbonney: I don't believe it can be source_generic or source_data, as the URN was removed from source_generic, or am I misunderstanding and you'd prefer it remained a 'subclass' of that type? The same can be said for the existing source_audio.

garethsb: I also brought up the point of data-binding JSON bodies to polymorphic types. However, it's not entirely possible already, hence a few instances of `"anyOf"` rather than `"oneOf"` in the schemas already. For what it's worth, I think the IS-05 v1.1 `/transporttype` solves your problem there (albeit in a different endpoint).

I don't think there is an overlap for `"format": "urn:x-nmos:format:data"` though? That is source_data.json not source_generic.json, according to this PR.

I made a related comment as to the applicability of `"event_type"` to other specs here - https://github.com/bbc/nmos-discovery-registration/pull/7#issuecomment-498330172.

mjeras: I apologize, didn't notice the removal of data from source_generic.json

garethsb:
A discussion on the group call today was whether `"event_type(s)"` are theoretically useful for several specifications working with `"urn:x-nmos:format:data"`, and therefore we could get rid of specific references to IS-07 in the schemas.

andrewbonney:
I'll remove the references for now. If these are used elsewhere it will be important to avoid duplicating types for different purposes. I'd suggest when we get chance that these attributes (and others already in IS-04) are noted under Source and Flow headings in the parameter registers so that we have somewhere a little more agnostic to point at other specifications like IS-07 from.

andrewbonney:
@mjeras I've removed IS-07 mentions from the schemas. The examples still mention it at the moment, but let me know if you think they should be modified too. Otherwise we'll merge this into the v1.3-dev branch.

andrewbonney:
Merging this now. If there are any further comments please add a new issue. We can still deal with that in review of the v1.3-dev branch ahead of release.

PR #105, opened 2019-06-14
docs: update data model diagram and referential integrity notes

### LINKED ISSUE #72: Improve documentation around referential integrity between versions
Following on from an e-mail discussion with @garethsb-sony 

* Document reasoning behind the referential integrity requirements and in particular the change between v1.0 and v1.1 relating to content transformers.
* Update the Data Model diagram to reflect the association between a Flow and Device for referential integrity purposes (note that this is scoped within IS-04, there is no requirement for a Flow to be attached to a single Device outside of an IS-04 registration)
* Indicate in the schema descriptions which attributes serve as super-resources with regard to referential integrity.

andrewbonney:
Resolves #72 

@garethsb-sony I can't remember whether the first bullet in the issue still needed dealing with or whether it's already covered by https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.2.x/docs/4.1.%20Behaviour%20-%20Registration.md#referential-integrity


garethsb:
LGTM. I agree that text covers the issue's first bullet.

PR #107, opened 2019-06-17
docs: add pagination examples covering edge cases

### LINKED ISSUE #21: Improve documentation around edge cases for pagination
This was originally requested via AMWA Basecamp (text below). Additions should be made to https://github.com/AMWA-TV/nmos-discovery-registration/blob/master/docs/2.5.%20APIs%20-%20Query%20Parameters.md#pagination

- What should the X-Paging-Since value be when a client specifies only a paging.until value, and it is before the oldest resource's timestamp?
The X-Paging-Until value should be the requested value, and the body should be an empty JSON array, but what about the X-Paging-Since value?  Should it be the epoch (i.e. "0:0"), or one nanosecond before X-Paging-Since, or simply omitted?

- Conversely, what should the X-Paging-Until value be when a client specifies only a paging.since value, and it is after the newest resource's timestamp?

- The client specifies a filter query that results in a single value, but no paging query parameters - what should the X-Paging-Since and X-Paging-Until values be?
Presumably the X-Paging-Until value is the timestamp of the returned resource, but what of X-Paging-Since?

- As above, but the filter query returns no results?

andrewbonney:
Resolves #21

### DIFF
--- a/docs/2.5. APIs - Query Parameters.md
+++ b/docs/2.5. APIs - Query Parameters.md
@@ -20,7 +20,7 @@ The following implementation notes should be observed:
 
 * The choice to page based on 'creation' or 'update' timestamp depends on the client's intended use of the data. Paging based on 'creation' time from the start of a collection provides the best mechanism to identify all resources held in a registry. Paging based on 'update' time from the end of a collection provides the best mechanism to watch for changing resources in the case that Websocket subscriptions are not available.
 
-* Paging through a collection for a second time, using the same cursors as previously should never result in new data appearing in the response payloads when using 'creation' ordering. However, note that as the registry is dynamic and resources may be updated or deleted a given cursor may return fewer results than it did previously. When using 'update' ordering and a query filter, new data may appear when cursors are used for a second time.
+* Paging through a collection for a second time, using the same cursors as previously should never result in new data appearing in the response payloads when using 'creation' ordering, provided the requested paging times are less than or equal to the maximum paging time held in the registry. However, note that as the registry is dynamic and resources may be updated or deleted a given cursor may return fewer results than it did previously. When using 'update' ordering and a query filter, new data may appear when cursors are used for a second time.
 
 * When query parameters which perform filtering are used at the same time as paging, the filters must be applied by the implementation before applying paging parameters to the resulting data set.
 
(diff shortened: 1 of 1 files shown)


docs/2.5. APIs - Query Parameters.md
```diff
@@ -216,6 +216,113 @@ Payload Resources
...
+GET /x-nmos/query/v1.1/nodes?paging.until=0:20
+```
+
+In this case, assume that there are stored records for '0:21' and '0:22' but no earlier.
+
+***Response***
+
+Headers
+```
+Link: <http://example.api.com/x-nmos/query/v1.1/nodes/?paging.since=0:20&paging.limit=10>; rel="next"
```
garethsb: Do we need to indicate that the `paging.limit=10` and `X-Paging-Limit: 10` are set in these examples based on the server's default?

garethsb: Do we need to indicate that server may (should?) also include `"prev"`, `"first"` and `"last"` even in this case (and also similarly all four in the other examples)?

docs/2.5. APIs - Query Parameters.md
```diff
@@ -20,7 +20,7 @@ The following implementation notes should be observed:
 
 * The choice to page based on 'creation' or 'update' timestamp depends on the client's intended use of the data. Paging based on 'creation' time from the start of a collection provides the best mechanism to identify all resources held in a registry. Paging based on 'update' time from the end of a collection provides the best mechanism to watch for changing resources in the case that Websocket subscriptions are not available.
 
-* Paging through a collection for a second time, using the same cursors as previously should never result in new data appearing in the response payloads. However, note that as the registry is dynamic and resources may be updated or deleted a given cursor may return fewer results than it did previously.
+* Paging through a collection for a second time, using the same cursors as previously should never result in new data appearing in the response payloads when using 'creation' ordering. However, note that as the registry is dynamic and resources may be updated or deleted a given cursor may return fewer results than it did previously. When using 'update' ordering and a query filter, new data may appear when cursors are used for a second time.
```
garethsb: Does the (contrived?) example of sending a request anchored to a future time like `paging.until=2345678901:0` need calling out as possibly returning new resources, or does "using the same cursors as previously" make clear its the returned `X-Paging-Since/Until` values that are being referred to?

docs/2.5. APIs - Query Parameters.md
andrewbonney: I've added something which matches this up to a note against example 1.

garethsb (approved):

PR #109, opened 2019-06-17
Add signalling for authorization requirements relating to BCP-003-02

### LINKED ISSUE #101: Consider any attribute or TXT record additions required to support BCP-003-02
BCP-003-02 adds a requirement to pass Authorization headers in requests. These could be required to read/write IS-04 resources on Nodes/Registries, or could be required to interact with services/controls which are advertised from IS-04. The requirement to include an Authorization token may be deployment specific, and may only be required for certain aspects of a deployment.

We should consider whether this should be flagged via DNS-SD TXT records prior to interacting with the APIs (may not be possible for the Node API directly), and whether an attribute should be added to service/control advertisements to indicate any Authorization requirements. The alternative is that this is pre-configured across a deployment in a static fashion, but this may limit flexibility.

andrewbonney:
Resolves #101 

I've intentionally made none of these attributes 'required' and set the default to 'false'. Where authorization is enabled they will need to be present and set to 'true'.

I'm not sure it's really necessary to permit inclusion of the attribute in Query API WebSocket requests, but I have done in case there is a genuine use case where this is useful. The notes reflect the general case where Query APIs will be operating in one mode only.

### DIFF
--- a/APIs/schemas/device.json
+++ b/APIs/schemas/device.json
@@ -67,6 +67,11 @@
                 "type": "string",
                 "description": "URN identifying the control format",
                 "format": "uri"
+              },
+              "authorization": {
+                "type": "boolean",
+                "description": "This endpoint requires authorization",
+                "default": false
               }
             }
           }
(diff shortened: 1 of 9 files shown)


garethsb:
LGTM.

PR #110, opened 2019-06-18
docs: re-order upgrade advice to avoid issues with pre-upgraded Nodes

andrewbonney:
There is a chance that if a Node in the system already supports v1.3 ahead of a system upgrade from v1.2 it may start registering at v1.3 once the new Registration APIs become available. If the downgrade supporting clients are upgraded first this should be less of an issue as these Nodes will always remain visible.


garethsb:
LGTM.

Issue #98, opened 2019-05-14
Reconsider automatic Query API key stripping functionality


garethsb:
@andrewbonney said that https://github.com/AMWA-TV/nmos-event-tally/issues/38 adds "further weight to a feeling I've had that we ought to be removing the automatic Query API key stripping functionality and depending upon the query downgrade mechanism for multi-version support instead, but that's very much an IS-04 discussion point."

@garethsb-sony "So v1.3 resources would never appear in a response to a v1.2 API?"

@andrewbonney "Yes, that's the thinking. You could as a result only cause mixed version responses by using /v1.3/resources/?query.downgrade=v1.2 etc, ensuring that only clients which are explicitly written to support multiple versions are exposed to it."

@garethsb-sony "That makes sense but loses us some of the flexibility discussed in IS-04 [Upgrade Path - Performing Upgrades](https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.2.x/docs/6.0.%20Upgrade%20Path.md#performing-upgrades)."

andrewbonney:
Just to add a few more notes for consideration. We first encountered an issue like this with the ability to set a Sender 'flow_id' to 'null' introduced in v1.1. This meant that when viewing this record via a Query API set to v1.0, a v1.0 client may encounter these values which technically breach the v1.0 schema. A similar requirement is now appearing in https://github.com/AMWA-TV/nmos-discovery-registration/pull/97 which will cause the same problem for v1.0-v1.2 clients when working against a registry which can also handle v1.3.

The cases above are bad because they are potentially unforeseeable changes from a client's perspective. There is also a further class of additions which could cause issues for legacy clients. This includes the addition of support for new transport types. Whilst clients should be checking the 'transport' key, as v1.0 to v1.2 have primarily been used for RTP some assumptions may have been made.

The aim of the key stripping functionality was to ensure that legacy clients could continue to function in upgraded systems. The danger is that whilst they may continue to function for a while, as more new Nodes are introduced to the system there is a greater chance that they may suddenly fail. The trade-off is of course that by changing this mechanism, the legacy clients would only be able to communicate with Nodes which continue to operate at the lower version(s).

The upgrade path does already suggest upgrading clients to the new version before upgrading Nodes (which protects against this issue), but there is no guarantee that this guidance will be followed in practice.

andrewbonney:
[...]

PR #112, opened 2019-06-18
docs: add a note regarding caps, sources and flows and potential future uses

### LINKED ISSUE #89: Expand defined set of Receiver capabilities
At present Receiver capabilities are defined to differentiate between the different media_types which Flows may use. In order for control systems to make informed routing decisions these need to cover a greater spectrum of Source/Flow attributes. Some initial prototyping in this area has been carried out in https://github.com/bbc/nmos-discovery-registration/pull/6, but is incomplete.

andrewbonney:
This is intended to resolve #89 as far as v1.3 goes. It doesn't make any changes, but notes explicitly that additional 'caps' may be defined externally in the future in order to avoid unnecessary version bumps.

'caps' have been omitted from the version translations documentation up to now, so wouldn't necessarily need explicit handling in relation to visibility across the boundaries of API versions.

### DIFF
--- a/docs/4.3. Behaviour - Nodes.md
+++ b/docs/4.3. Behaviour - Nodes.md
@@ -6,6 +6,10 @@ _(c) AMWA 2016, CC Attribution-ShareAlike 4.0 International (CC BY-SA 4.0)_
 
 Where the behaviour associated with an API attribute is not sufficiently clear from its name and schema constraints, this is documented below.
 
+### All Resources
+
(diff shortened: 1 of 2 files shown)


andrewbonney:
I've added a commit relating to Source and Flow attributes here too. The final commit which adjusts the version translation text is intended to be backported to v1.2.x and v1.1.x. Together this is intended to make #98 less critical for the upcoming release.

PR #115, opened 2019-07-11
docs: prevent mDNS browsing when unicast records are available

andrewbonney:
This PR adjusts the discovery procedure to ensure that when unicast records are available, these are used in preference to any mDNS records which may be available on the same network. This holds true unless the Node supports a configuration parameter for the user to choose between unicast and multicast DNS.


garethsb:
LGTM

PR #118, opened 2019-08-01
node: add explicit 409 code for version conflicts

garethsb:
Follow up to #85.

I'm still unsure about using the same UUID for all resources in the example Location headers.
And I'd prefer the URLs to end without a trailing slash in all cases.

But for the moment, I've gone for consistency with the Query API examples.


garethsb:
Now updated to match all UUIDs in Location header examples to the relevant example response data, and removed trailing slashes as this demonstrates what a client must use to make subsequent DELETE requests on the Registration API.

PR #130, opened 2019-10-10
Fixing example inaccuracies

### LINKED ISSUE #129: Example Inaccuracies
I have noticed two issues while writing some unit tests using the examples provided in this repository:
- nodeapi-flows-get-200.json - The IS-07 flows (as opposed to the other flows in the array) have "caps" fields, although they are not defined for flows. That seems to originate from the IS-07 documentation samples and has to be corrected there too. The fix is to remove the fields.
- registrationapi-resource-*.json (3 files affected) - The examples show the update for a node, but the node in the example doesn't have "interfaces" property which is mandatory for a node. The fix is to (at least) add an empty "interfaces" array.
I have the fixes for the above locally, but don't have the rights to push them and create a pull request.

mjeras:
Resolves #129


garethsb (changes requested):
Removing `"caps"` is perfect.

I think it would be better to make the Node in the registrationapi-*.json files back in sync with the one at https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.3.x/examples/nodeapi-self-get-200.json, though in all these places the `"api"` `"versions"` needs `"v1.2"` and `"v1.3"` adding!

garethsb:
Just noting that having JSON Schema validation in the continuous integration scripts (https://github.com/AMWA-TV/nmos-doc-build-scripts/issues/1) would allow us to verify this PR.

peterbrightwell:
Fixed in #156

PR #138, opened 2019-11-27
Fix NSCnnn and Unn regexes

garethsb:
The channel symbol regexes for Numbered Source Channel and Undefined are too lax.

Before this fix, the following would have been accepted:

* fooNSC001bar
* fooU01bar
* NSC000

Verified with the help of https://regex101.com/.

### DIFF
--- a/APIs/schemas/source_audio.json
+++ b/APIs/schemas/source_audio.json
@@ -67,11 +67,11 @@
                     ]
                   },
                   {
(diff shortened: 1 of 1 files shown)


garethsb:
> 1 through 127 does not allow for the 128th track, the previous version intended to started counting from 0 which allowed for the even number of tracks.

VSF TR-03 is the source specification, and it says: "NSC[CHNUM] - Numbered Source Channel [CHNUM] . A single channel of audio that is ordered between 001 and 127 and is intended to play out in a Discrete Numbered Sources Soundfield Group; for example NSC025."

PR #140, opened 2019-12-12
Fixed NSC regex

### LINKED ISSUE #139: Minor error in regex for source_audio schema
The current regex for Numbered Source Channel has a small error, currently only allowing NSC001-127. 

This should be NSC001-NSC128 (or NSC000-NSC127).

As 'Undefined Channel' ranges from U01-U64 I assume we want the first version (NSC001-NSC128).

jasperpeeters:
Fixes #139


jasperpeeters:
Due to spec NSC001-NSC127 is correct.

Issue #149, opened 2020-03-17
Query API WebSocket upgrade


garethsb:
On the one hand (in [Creating a WebSocket subscription](https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.3.x/docs/4.2.%20Behaviour%20-%20Querying.md#creating-a-websocket-subscription)):

> Connection upgrade procedures permitting an existing HTTP query to transition to a websocket are not currently supported, but may be in the future.

On the other hand (one paragraph further on, in [Query API Behaviour](https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.3.x/docs/4.2.%20Behaviour%20-%20Querying.md#query-api-behaviour)):

> Query APIs may additionally support the HTTP 'Upgrade' header sent by clients to upgrade an HTTP GET request to a Websocket.

garethsb:
We discussed one fix, to remove the first sentence and clarify the second one:

Query APIs may support the HTTP protocol upgrade mechanism on the list resources like `/nodes` to transition an HTTP GET request to a WebSocket connection, but are not required to do so.

Could link to e.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism or https://tools.ietf.org/html/rfc7230#section-6.7 and https://tools.ietf.org/html/rfc6455?

Presumably the created subscription will be non-persistent?

garethsb:
More guidance, with the link to the MDN documentation mentioned above, could be added to a future implementation guide in the style of INFO-002.

Issue #20, opened 2017-08-31
Mandate support for a core set of RQL operators when implemented in a Query API


andrewbonney:
At present the Query API spec provides a recommended set of operators to implement when using the optional RQL support in the spec. It has been suggested that this is strengthened to be a mandatory set of operators which must be present if RQL support is implemented.

This is being tracked on the AMWA Basecamp but is being recorded here for completeness.

andrewbonney:
As part of this update it has been noted that the currently suggested 'sort()' operator could cause conflicts with the paging mechanism, causing it to be non-deterministic. This may need adding to the list of operators which are highlighted to explicitly not be implemented.

andrewbonney:
Should also consider use of 'select()' to reduce the fields returned in cases where clients don't require them, and potentially a form of 'rel()' to search within foreign keys, for example:
```
/flows?query.rql=rel(senders,matches(transport,urn%3Ax-nmos%3Atransport%3Artp))
```

garethsb:
See https://github.com/AMWA-TV/nmos-receiver-capabilities/issues/16 for a demonstration of how RQL can be used to implement a useful range of complex queries based on proposed BCP-004-01 Receiver Capabilities to identify the matching Senders. They require that the Query API implement:
* the appropriate Query API-specific semantics for RQL _property names_ like 'label' and, using the '.' syntax, like 'api.endpoints.host' for querying in objects and arrays (as already recommended in [Advanced (RQL) Queries](https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.3.x/docs/2.5.%20APIs%20-%20Query%20Parameters.md#advanced-rql-queries-optional) to match the Basic Queries syntax)
* a core set of RQL _call-operators_ including `and`, `or`, `in`, `eq`, `ge`, `le`, `count`, and `rel` with the appropriate Query API-specific semantics to use identifier property names like 'flow_id' as _relation names_ (cf. the suggested list of operators in the spec, [Advanced (RQL) Queries - Constraints](https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.3.x/docs/2.5.%20APIs%20-%20Query%20Parameters.md#constraints)) 
* RQL _explicitly-typed values_ for a couple of NMOS types, e.g. `rational:30000%2f1001` for 30000/1001

peterbrightwell:
This issue has been labelled as an enhancement and captured on the internal roadmap which will determine when resources can be allocated to address it.

PR #158, opened 2021-04-01
Normative language improvements

peterbrightwell:
Normative terms (MUST, SHOULD, MAY, etc.) are now all capitalised. Non-normative uses of those terms (which were mostly lower-case) have been replaced or rephrased.

Some other language improvements also included and a little reformatting.


docs/1.0. Overview.md
```diff
@@ -1,21 +1,34 @@
...
 
-The documents included in this directory provide additional details and recommendations for implementations of the defined APIs, or their consumers.
+## Introduction
 
-Familiarity with the [JT-NM reference architecture](https://jt-nm.org/reference-architecture/) is assumed, but a summary of the resources referenced by this specification is available in the [Data Model](5.0.%20Data%20Model.md).
+AMWA IS-04 specifies how to register and discover resources in an NMOS system. IS-04 is built on three APIs:
 
-## Introduction
+- The [Registration API](../APIs/RegistrationAPI.raml) allows a Node to register its resources.
+- The [Query API](../APIs/QueryAPI.raml) allows query of registered resources.
```
garethsb:
```suggestion
- The [Query API](../APIs/QueryAPI.raml) allows querying of registered resources.
```

docs/1.0. Overview.md
```diff
@@ -1,21 +1,34 @@
...
+- The [Node API](../APIs/NodeAPI.raml) is used to find resources on a Node, and is used for [peer-to-peer discovery](3.2.%20Discovery%20-%20Peer%20to%20Peer%20Operation.md)
+
+The Specification includes:
 
-The purpose of this document is to explain how the individual entities in an NMOS system can be made discoverable.
+- RAML and JSON Schema definitions, with supporting JSON examples
+- This documentation set, which provides:
+  - An overview of the API and how it is used.
```
garethsb:
```suggestion
  - An overview of each API and how it is used.
```

docs/2.0. APIs.md
```diff
@@ -27,7 +29,7 @@ All APIs MUST provide a JSON representation signalled via 'Content-Type: applica
 
 All NMOS APIs MUST use a path in the following format. Other HTTP resources MAY be presented on the same port by the Node, provided all NMOS resources are available from the /x-nmos/ path as follows:
 
-```
+```json
```
garethsb: This isn't a JSON snippet?

```suggestion
```
```

File: docs/2.0. APIs.md [hunk 4]
```diff
@@ -41,33 +43,34 @@ Where protocol and versioning data is not available, clients MAY assume a v1.0 A
...
-* Appending /v1.0/ to the API base resource will request version 1.0 of the API if available.
-* The versioning format is v&lt;#MAJOR&gt;.&lt;#MINOR&gt;
-* MINOR increments SHOULD be performed for non-breaking changes (such as the addition of attributes in a response)
-* MAJOR increments MUST be performed for breaking changes (such as the renaming of a resource or attribute)
-* Versions MUST be represented as complete strings. Parsing MUST proceed as follows: separate into two strings, using the point (.) as a delimiter. Compare integer representations of MAJOR, MINOR version (such that v1.12 is greater than v1.5).
-* Clients are responsible for identifying the correct API version they require.
+- Appending /v1.0/ to the API base resource will request version 1.0 of the API if available.
+- The versioning format is v&lt;#MAJOR&gt;.&lt;#MINOR&gt;
+- MINOR increments SHOULD be performed for non-breaking changes (such as the addition of attributes in a response)
+- MAJOR increments MUST be performed for breaking changes (such as the renaming of a resource or attribute)
```
garethsb: These requirements are on specification maintainers not implementers, so as we've discussed previously, these could be rephrased as:

```suggestion
- MINOR increments will be performed for non-breaking changes (such as the addition of attributes in a response)
- MAJOR increments will be performed for breaking changes (such as the renaming of a resource or attribute)
```

This language is also repeated in other specs...

docs/2.0. APIs.md
```diff
@@ -77,34 +80,37 @@ In order to overcome shortcomings in some common libraries, the following requir
...
+- Clients performing requests using these methods MUST use URLs with no trailing slash present.
+- Servers implementing the APIs MUST correctly interpret requests using these methods to paths without trailing slashes present.
+- Servers implementing the APIs MAY correctly interpret requests using these methods to paths with trailing slashes present for backward compatibility.
+- Servers SHOULD NOT respond with 3xx codes for these request types.
 
 #### Client Behaviour
 
-When a server implementation is required to indicate an API URL via an 'href' or similar attribute, it is valid to either include a trailing slash or not provided that the listed path is accessible and abides by the policy above. In order to avoid cases of double slashes or missing slashes, clients are required to exercise caution when appending paths onto 'href' type attributes.
+When a server implementation needs to indicate an API URL via an 'href' or similar attribute, it is valid to either include a trailing slash or not provided that the listed path is accessible and follows the above rules. Clients appending paths to 'href' type attributes MUST support both cases, avoiding doubled or missing slashes.
```
garethsb: Just noting that I think this language is repeated in other specs.

docs/2.0. APIs.md
```diff
@@ -77,34 +80,37 @@ In order to overcome shortcomings in some common libraries, the following requir
...
+```json
 {
   "code": 400,
   "error": "Human readable message which is suitable for user interface display, and helpful to the user",
   "debug": "Programmer / debugging detail or traceback"
 }
 ```
 
-In the above example, the 'code' should always match the HTTP status code. 'error' must always be present and in string format. 'debug' may be null if no further debug information is available.
+In the above example, the 'code' SHOULD match the HTTP status code. 'error' MUST always be present and in string format. 'debug' MAY be null if no further debug information is available.
```
```suggestion
The 'code' SHOULD match the HTTP status code. 'error' MUST always be present and in string format. 'debug' MAY be null if no further debug information is available.
```

docs/2.1. APIs - Common Keys.md
```diff
@@ -83,20 +84,21 @@ See [Use of URNs](#use-of-urns) below for additional requirements regarding thei
...
+- The URN MAY include a version number
 
 Subclassifications are defined as the portion of the URN which follows the first occurrence of a '.', but prior to any '/' character. By convention, URNs which begin 'urn:x-nmos:' will never include a further ':' after the first occurrence of a '.'.
 
 For example, the URN base 'urn:x-nmos:transport:rtp' above has two subclassifications of 'ucast' and 'mcast'.
 
 API clients interpreting URNs MUST correctly recognise any as-yet undefined subclassifications as being associated with the URN base which appears prior to the first '.' character.
 
-For example, if filtering by the format 'urn:x-nmos:format:video', the filter is required to include any as-yet undefined subclassifications such as 'urn:x-nmos:format:video.raw'.
+For example, if filtering by the format 'urn:x-nmos:format:video', the filter includes any as-yet undefined subclassifications such as 'urn:x-nmos:format:video.raw'.
```
garethsb: Just noting that in the long time since this was originally written, practice has shown that we don't need to add such subclassifications, we use other mechanisms, such as IANA `media_type` in this example, so this might be something we want to change in any v1.4 or v2.0.

docs/2.3. APIs - Server Side Implementation Notes.md
```diff
@@ -8,9 +8,9 @@ In order to permit web-based control interfaces to be hosted remotely, all NMOS
 
 In addition to this, where highlighted in the API specifications servers MUST respond to HTTP pre-flight OPTIONS requests. Servers MAY additionally support HTTP OPTIONS requests made to any other API resource.
 
-In order to simplify development, the following headers may be returned in order to remove these restrictions as far as possible. Note that these are very relaxed and may not be suitable for a production deployment.
+In order to simplify development, the following headers MAY be returned in order to remove these restrictions as far as possible. Note that these are very relaxed and possibly not suitable for a production deployment.
 
-```
+```json
```
garethsb: This example isn't JSON. I think this would work better...

```suggestion
```http
```

docs/2.0. APIs.md
```diff
@@ -3,21 +3,23 @@
...
+
+- The following sub-sections describing common API properties.
+- [RAML](http://raml.org/) documents and [JSON schemas](http://tools.ietf.org/html/draft-zyp-json-schema-04) in the [APIs](../APIs/) folder.
 
 Examples of JSON format output are provided in the [examples](../examples/) folder.
 
 ## API Validation
 
-JSON schemas are included with the RAML API definitions. These include validation for values used within the APIs. It is strongly recommended that implementers of a Registration API use these JSON schemas as part of a validation stage when receiving registrations from Nodes. Invalid resources should cause a 400 HTTP error (Bad Request) to be returned to the client.
+JSON schemas are included with the RAML API definitions. These include validation for values used within the APIs. It is RECOMMENDED that implementers of a Registration API use these JSON schemas as part of a validation stage when receiving registrations from Nodes. Invalid resources SHOULD cause a 400 HTTP error (Bad Request) to be returned to the client.
```
garethsb: Reinstate strongly?

```suggestion
JSON schemas are included with the RAML API definitions. These include validation for values used within the APIs. It is strongly RECOMMENDED that implementers of a Registration API use these JSON schemas as part of a validation stage when receiving registrations from Nodes. Invalid resources SHOULD cause a 400 HTTP error (Bad Request) to be returned to the client.
```

docs/2.5. APIs - Query Parameters.md
```diff
@@ -12,60 +12,65 @@ Query APIs SHOULD support pagination of their API resources where the 'paged' tr
 
 Query API clients MUST detect whether pagination is being used by examining the HTTP response headers for 'X-Paging-Limit' which MUST be returned in all cases where pagination is in use.
 
-The following implementation notes should be observed:
+The registry which backs the Query API SHOULD maintain a 'creation' and 'update' timestamp alongside each registered resource. These values SHOULD NOT be returned to API clients in the response body, but will be made available via headers and used as pagination cursors.
 
-* The registry which backs the Query API is expected to maintain a 'creation' and 'update' timestamp alongside each registered resource. These values should not be returned to API clients in the response body, but will be made available via headers and used as pagination cursors.
+To ensure that pagination does not result in resources being skipped, there SHOULD NOT be duplicate creation or update timestamps stored against resources of the same type (Node, Source, Flow etc). It is RECOMMENDED that these timestamps are stored with nanosecond resolution using a TAI timebase, which will allow clients to navigate collections based on a common understanding of time.
```
garethsb: While we're here...

```suggestion
To ensure that pagination does not result in resources being skipped, there SHOULD NOT be duplicate creation or update timestamps stored against resources of the same type (Node, Source, Flow, etc.). It is RECOMMENDED that these timestamps are stored with nanosecond resolution using a TAI timebase, which will allow clients to navigate collections based on a common understanding of time.
```

docs/2.5. APIs - Query Parameters.md
```diff
@@ -12,60 +12,65 @@ Query APIs SHOULD support pagination of their API resources where the 'paged' tr
...
 
-* In order to ensure that pagination does not result in resources being skipped, it is important that there are no duplicate creation or update timestamps stored against resources of the same type (Node, Source, Flow etc). It is suggested (although not mandated) that these timestamps are stored with nanosecond resolution using a TAI timebase, which will allow clients to navigate collections based on a common understanding of time.
+The choice to page based on 'creation' or 'update' timestamp SHOULD depend on the client's intended use of the data:
```
garethsb: Is this stating a recommendation?

docs/2.5. APIs - Query Parameters.md
```diff
@@ -12,60 +12,65 @@ Query APIs SHOULD support pagination of their API resources where the 'paged' tr
...
 
-* Paging through a collection for a second time, using the same cursors as previously should never result in new data appearing in the response payloads when using 'creation' ordering, provided the requested paging times are less than or equal to the maximum paging time held in the registry. However, note that as the registry is dynamic and resources may be updated or deleted a given cursor may return fewer results than it did previously. When using 'update' ordering and a query filter, new data may appear when cursors are used for a second time.
+Paging through a collection for a second time, using the same cursors as previously SHOULD NOT result in new data appearing in the response payloads when using 'creation' ordering, provided the requested paging times are less than or equal to the maximum paging time held in the registry. However, note that as the registry is dynamic and resources MAY be updated or deleted a given cursor MAY return fewer results than it did previously. When using 'update' ordering and a query filter, new data MAY appear when cursors are used for a second time.
```
garethsb: This is a little awkward because the sentence seems to begin from the point of view of the _client_, but the "SHOULD NOT" as a recommendation is addressing the _server_.

The instances of "may" in the second sentence were also really stating fact from the point of view of the client; as conformance language most of them apply to the server, so again, a bit awkward.

```suggestion
Paging through a collection for a second time, using the same cursors as previously SHOULD NOT result in new data appearing in the response payloads when using 'creation' ordering, provided the requested paging times are less than or equal to the maximum paging time held in the registry. However, note that as the registry is dynamic and resources could be updated or deleted a given cursor MAY return fewer results than it did previously. When using 'update' ordering and a query filter, new data MAY appear when cursors are used for a second time.
```

docs/2.5. APIs - Query Parameters.md
```diff
@@ -12,60 +12,65 @@ Query APIs SHOULD support pagination of their API resources where the 'paged' tr
...
+When query parameters which perform filtering are used at the same time as paging, the filters MUST be applied by the implementation before applying paging parameters to the resulting data set.
 
-* A server may choose its own default value for paging limit (see Example 1).
+A server MAY choose its own default value for paging limit (see Example 1).
 
-* Where both 'since' and 'until' parameters are specified, the 'since' value takes precedence where a resulting data set is constrained by the server's value of 'limit' (see Example 5).
+Where both 'since' and 'until' parameters are specified, the 'since' value takes precedence where a resulting data set is constrained by the server's value of 'limit' (see Example 5).
 
-* Servers should include 'prev' and 'next' Link headers and may include 'first' and 'last' Link headers in all responses.
+Servers SHOULD include 'prev' and 'next' Link headers and MAY include 'first' and 'last' Link headers in all responses.
```
garethsb: 'first' and 'last' are optional but dead easy... first = `paging.since=0:0`, last = omit `paging.since` and `paging.until`

docs/2.5. APIs - Query Parameters.md
```diff
@@ -12,60 +12,65 @@ Query APIs SHOULD support pagination of their API resources where the 'paged' tr
...
 
-The following examples aim to identify how pagination should behave in the presence of a set of registered data. In order to avoid displaying full resource representations, the only data listed here is the 'update' timestamp associated with each registered record. The same procedures can be applied where 'creation' timestamps and the '?paging.order=create' parameter are used instead.
+The following examples show pagination for a set of registered data. In order to avoid displaying full resource representations, the only data listed here is the 'update' timestamp associated with each registered record. The same procedures can be applied where 'creation' timestamps and the '?paging.order=create' parameter are used instead.
 
 Where a paging limit is not specified in a request the server's default is used.
 
 **Sample Data: Registered Node Update Timestamps (Comma-Separated)**
-```
+
+```json
```
garethsb, line 42: This example is not JSON (even though it has `[` and `]` to indicate an array).

Given that the suggestion would be to just to go back to three back-ticks, it's impossible to write!!

Or quote the timestamps?
Or accept that it's not valid JSON?
Hmmm...

docs/2.5. APIs - Query Parameters.md
```diff
@@ -12,60 +12,65 @@ Query APIs SHOULD support pagination of their API resources where the 'paged' tr
...
 [0:1, 0:2, 0:3, 0:4, 0:5, 0:6, 0:7, 0:8, 0:9, 0:10, 0:11, 0:12, 0:13, 0:14, 0:15, 0:16, 0:17, 0:18, 0:19, 0:20]
 ```
 
 Each of the above corresponds to the update timestamp of a corresponding Node, in the format &lt;seconds&gt;:&lt;nanoseconds&gt; and displayed in ascending order. These will be used throughout the following examples.
 
-Response payloads in the examples will show these values, but in a real implementation should be replaced by the corresponding JSON objects for the Nodes or other resources being queried.
+Response payloads in the examples will show these values, but in a real implementation SHOULD be replaced by the corresponding JSON objects for the Nodes or other resources being queried.
```
garethsb: If this were conformance language it would have to be MUST, which would look like a substantive change, so I suggest:

```suggestion
Response payloads in the examples will show these values, but in a real implementation would be replaced by the corresponding JSON objects for the Nodes or other resources being queried.
```

docs/2.5. APIs - Query Parameters.md
```diff
@@ -12,60 +12,65 @@ Query APIs SHOULD support pagination of their API resources where the 'paged' tr
...
 Link: <http://example.api.com/x-nmos/query/v1.1/nodes/?paging.since=0:20&paging.limit=10>; rel="next", <http://example.api.com/x-nmos/query/v1.1/nodes/?paging.until=0:10&paging.limit=10>; rel="prev"
 X-Paging-Limit: 10
 X-Paging-Since: 0:10
 X-Paging-Until: 0:20
 ```
 
 Payload Resources
-```
+
+```json
```
garethsb, line 73: Again, not JSON, but should have same resolution as above.

docs/2.5. APIs - Query Parameters.md
```diff
@@ -81,31 +86,35 @@ Payload Resources
...
-* The Link header identifies the 'next' and 'prev' cursors which an application may use to make its next requests. An implementation may also provide 'first' and 'last' cursors to identify the beginning and end of a collection if this can be addressed via a consistent URL.
+
+- The data set returned when no query parameters are specified MUST be from the most recently updated (or created) resources in the collection, returned in descending order.
+- The X-Paging headers identify properties of the collection of data returned in the response. These parameters MAY be used to construct a URL which would return the same set of bounded data on consecutive requests (for example /x-nmos/query/v1.1/nodes?paging.since=0:10&paging.until=0:20).
+- The Link header identifies the 'next' and 'prev' cursors which an application MAY use to make its next requests. An implementation MAY also provide 'first' and 'last' cursors to identify the beginning and end of a collection if this can be addressed via a consistent URL.
```
garethsb: Not sure if that first "MAY" is really a normative requirement, it might be better as "could" or "can"?

garethsb: As mentioned above, the 'first' and 'last' cursors can always be provided by consistent URLs.

docs/2.5. APIs - Query Parameters.md
```diff
@@ -81,31 +86,35 @@ Payload Resources
...
 Link: <http://example.api.com/x-nmos/query/v1.1/nodes/?paging.since=0:20&paging.limit=5>; rel="next", <http://example.api.com/x-nmos/query/v1.1/nodes/?paging.until=0:15&paging.limit=5>; rel="prev"
 X-Paging-Limit: 5
 X-Paging-Since: 0:15
 X-Paging-Until: 0:20
 ```
 
 Payload Resources
-```
+
+```json
```
garethsb, line 125: Same

docs/2.5. APIs - Query Parameters.md
```diff
@@ -116,27 +125,31 @@ Payload Resources
...
 Link: <http://example.api.com/x-nmos/query/v1.1/nodes/?paging.since=0:14&paging.limit=10>; rel="next", <http://example.api.com/x-nmos/query/v1.1/nodes/?paging.until=0:4&paging.limit=10>; rel="prev"
 X-Paging-Limit: 10
 X-Paging-Since: 0:4
 X-Paging-Until: 0:14
 ```
 
 Payload Resources
-```
+
+```http
```
garethsb: This should be `json` or whatever is consistent with the array response examples above.

docs/2.5. APIs - Query Parameters.md
```diff
@@ -218,18 +237,20 @@ Payload Resources
...
 
 ***Notes***
-* Whilst both 'since' and 'until' are specified, as this server example has a default paging limit of '10', the 'since' parameter must take precedence. As a result of this the value of X-Paging-Until is lower than requested and a further request must be made to retrieve any remaining data.
+
+- Whilst both 'since' and 'until' are specified, as this server example has a default paging limit of '10', the 'since' parameter  takes precedence. As a result of this the value of X-Paging-Until is lower than requested and a further request MUST be made to retrieve any remaining data.
 
 ### Edge Cases
 
-When a client requests data which falls at the extreme ends of the stored data set it may be less clear what values should be returned in the 'X-Paging-Limit' and 'X-Paging-Since' headers. The following examples are intended to clarify the expectation in these cases.
+When a client requests data which falls at the extreme ends of the stored data set it can be less clear what values SHOULD be returned in the 'X-Paging-Limit' and 'X-Paging-Since' headers. The following examples are intended to clarify these cases.
```
garethsb: This doesn't read like a requirement, how about:
```suggestion
When a client requests data which falls at the extreme ends of the stored data set it can be less clear what values ought to be returned in the 'X-Paging-Limit' and 'X-Paging-Since' headers. The following examples are intended to clarify these cases.
```

docs/2.5. APIs - Query Parameters.md
```diff
@@ -317,160 +346,180 @@ In this case, assume that the most recently created or updated Node held in the
...
-```
+
+```http
 GET /x-nmos/query/v3.0/flows?query.downgrade=v1.0
 ```
 
 ***Response***
-* Returns an HTTP 400 error code as downgrade queries may not be performed between major API versions.
+
+- Returns an HTTP 400 error code as downgrade queries MUST NOT be performed between major API versions.
```
garethsb: Agreed; "may not" meant "MUST NOT" :-)

docs/2.5. APIs - Query Parameters.md
```diff
@@ -317,160 +346,180 @@ In this case, assume that the most recently created or updated Node held in the
...
 
 ## Basic Queries
 
 Query APIs SHOULD support basic queries against their API resources. A 501 HTTP status code MUST be returned where a basic query is attempted against a Query API which does not implement it.
 
-Basic queries make use of standard HTTP GET query parameters in the form '?key1=value1&key2=value2'. Keys may match any attribute which the API schemas indicate could be returned by a given resource.
+Basic queries make use of standard HTTP GET query parameters in the form '?key1=value1&key2=value2'. Keys MAY match any attribute which the API schemas indicate could be returned by a given resource.
```
garethsb: This doesn't read quite right. This could be "MUST" (or "SHOULD" given the following sentence)? Or not use conformance language?

docs/2.5. APIs - Query Parameters.md
```diff
@@ -317,160 +346,180 @@ In this case, assume that the most recently created or updated Node held in the
...
 
 When an RQL query is specified, the Basic Queries format MAY be ignored in order to simplify implementation.
 
-Note that as RQL permits the definition of very complex queries, the server may return a 400 error code to indicate that it is refusing to action the client's request due to the query's complexity.
+Note that as RQL permits the definition of very complex queries, the server MAY return a 400 error code to indicate that it is refusing to action the client's request due to the query's complexity.
 
 ### Constraints
 
-Not all RQL operators will be suitable for use with a Query API. It is suggested that the following operators are supported, however these are not mandated and implementations may choose to support their own subset.
+Not all RQL operators will be suitable for use with a Query API. It is suggested that the following operators are supported, however these are not mandated and implementations MAY choose to support their own subset.
```
garethsb, line 540: FWIW, I'm thinking about proposing a BCP that defines a useful RQL subset much more tightly.

docs/2.5. APIs - Query Parameters.md
```diff
@@ -485,71 +534,83 @@ NB: It is suggested that sort() and limit() are not implemented as paging and as
...
+
+```http
 GET /x-nmos/query/v1.1/sources?query.rql=and(eq(format,urn%3Ax-nmos%3Aformat%3Avideo),in(tags.location,(Salford,London)))
 ```
 
 ***Response***
-* Returns all Sources which have a 'format' attribute equal to 'urn:x-nmos:format:video' AND a 'tags' attribute with a key of 'location', which has values of 'Salford' or 'London'.
 
-## Ancestry Queries (Optional)
+- Returns all Sources which have a 'format' attribute equal to 'urn:x-nmos:format:video' AND a 'tags' attribute with a key of 'location', which has values of 'Salford' or 'London'.
```
garethsb: Maybe we don't capitalize AND?

```suggestion
- Returns all Sources which have a 'format' attribute equal to 'urn:x-nmos:format:video' and a 'tags' attribute with a key of 'location', which has values of 'Salford' or 'London'.
```

docs/2.5. APIs - Query Parameters.md
```diff
@@ -485,71 +534,83 @@ NB: It is suggested that sort() and limit() are not implemented as paging and as
...
-* For each first generation parent found, their 'parents' array is used to identify the next generation of parents to be returned.
-* This process continues up to the generation limit returned in the X-Ancestry-Generations header (in this case limited to 2 by a query parameter). The server may refuse to honour a requested ''?query.ancestry_generations' value using a 400 error code if it is deemed to be too resource intensive.
+
+- Returns Flows which are listed in the 'parents' array of the Flow identified by the ID 'ad14888a-3a98-444c-8aa8-4d87b77cbaa1'. These are the first generation parents.
+- For each first generation parent found, their 'parents' array is used to identify the next generation of parents to be returned.
+- This process continues up to the generation limit returned in the X-Ancestry-Generations header (in this case limited to 2 by a query parameter). The server MAY refuse to honour a requested ''?query.ancestry_generations' value using a 400 error code if it is deemed to be too resource intensive.
```
garethsb: Is that repeated single quote intended?

docs/3.1. Discovery - Registered Operation.md
```diff
@@ -11,44 +11,51 @@ This document describes usage of NMOS APIs for discovery in cases where where a
...
+*Note: RFC6763 Section 7.2 specifies that the maximum service name length for an mDNS advertisement is 16 characters when including the leading underscore, but "_nmos-registration" is 18 characters.*
 
 The IP address and port of the Registration API MUST be identified via the DNS-SD advertisement, with the full HTTP path then being resolved via the standard NMOS API path documentation.
 
 Multiple DNS-SD advertisements for the same API are permitted where the API is exposed via multiple ports and/or protocols.
 
 ### DNS-SD TXT Records
 
-**api\_proto**
+### api\_proto
```
garethsb: Intended to indent this and subsequent TXT records one more, presumably?

docs/3.1. Discovery - Registered Operation.md
```diff
@@ -11,44 +11,51 @@ This document describes usage of NMOS APIs for discovery in cases where where a
...
 
 The DNS-SD advertisement MUST be accompanied by a TXT record of name 'api\_proto' with a value of either 'http' or 'https' dependent on the protocol in use by the Registration API web server.
 
-**api\_ver**
+### api\_ver
 
-The DNS-SD advertisement MUST be accompanied by a TXT record of name 'api\_ver'. The value of this TXT record is a comma separated list of API versions supported by the server. For example: 'v1.0,v1.1,v2.0'. There should be no whitespace between commas, and versions should be listed in ascending order.
+The DNS-SD advertisement MUST be accompanied by a TXT record of name 'api\_ver'. The value of this TXT record is a comma separated list of API versions supported by the server. For example: 'v1.0,v1.1,v2.0'. There SHOULD be no whitespace between commas, and versions SHOULD be listed in ascending order.
```
garethsb: These still ought to be MUST. Oh well.

docs/3.1. Discovery - Registered Operation.md
```diff
@@ -68,26 +75,31 @@ The DNS-SD advertisement MUST be accompanied by a TXT record of name 'api\_proto
 
 **api\_ver**
```
garethsb: This needs changing to use `####` like the previous ones, or those should go back to just bold.
I think that other spec documents also use bold rather than headings for TXT records, FWIW.

garethsb (commented):
Lots of good editorial changes, Pete. This is a partial review. I'll pick up from **docs/3.2. Discovery - Peer to Peer Operation.md** next...

docs/2.3. APIs - Server Side Implementation Notes.md
peterbrightwell: Yes you're right

docs/3.1. Discovery - Registered Operation.md
peterbrightwell: I agree but that would be beyond the scope of these changes.

docs/3.1. Discovery - Registered Operation.md
peterbrightwell: Quite likely yes. Will look later.

docs/3.2. Discovery - Peer to Peer Operation.md
```diff
@@ -2,18 +2,17 @@
...
+This document describes usage of NMOS APIs for discovery in cases where where a distributed registry is not available, such as small ad-hoc installations. This feature is OPTIONAL from v1.3 of the specification.
 
-Note that the peer-to-peer discovery mechanism depends upon mDNS which is not intended to operate over IP routed boundaries (ie. between network subnets). As such it is recommended that for most use cases, in particular fixed installations and those requiring high levels of resilience, the registered discovery mechanism is used.
+Note that the peer-to-peer discovery mechanism depends upon mDNS which is not intended to operate over IP routed boundaries (ie. between network subnets). As such it is RECOMMENDED that for most use cases, in particular fixed installations and those requiring high levels of resilience, the registered discovery mechanism is used.
 
 ## Pre-Requisites
 
-The following pre-requisites are optional aspects of the [Node API](../APIs/NodeAPI.raml) implementation which must be in place to permit peer-to-peer discovery.
+When configured for peer-to-peer discovery, a Nodes MUST advertise its presence wth mDNS.
```
garethsb:
```suggestion
When configured for peer-to-peer discovery, a Node MUST advertise its presence wth mDNS.
```

docs/3.2. Discovery - Peer to Peer Operation.md
```diff
@@ -23,23 +22,23 @@ The IP address and port of the Node API MUST be identified via the DNS-SD advert
...
+### api\_ver
 
-The DNS-SD advertisement MUST be accompanied by a TXT record of name 'api\_ver'. The value of this TXT record is a comma separated list of API versions supported by the server. For example: 'v1.0,v1.1,v2.0'. There should be no whitespace between commas, and versions should be listed in ascending order.
+The DNS-SD advertisement MUST be accompanied by a TXT record of name 'api\_ver'. The value of this TXT record is a comma separated list of API versions supported by the server. For example: 'v1.0,v1.1,v2.0'. There SHOULD be no whitespace between commas, and versions SHOULD be listed in ascending order.
 
-**api\_auth**
+### api\_auth
 
-The DNS-SD advertisement MUST be accompanied by a TXT record of name 'api\_auth' with a value of either 'true' or 'false' dependent on whether authorization is required in order to interact with the Node API or not.
+The DNS-SD advertisement MUST be accompanied by a TXT record of name 'api\_auth' with a value of either 'true' or 'false' dependent on whether authorization is needed for Registration API requests.
```
garethsb:
```suggestion
The DNS-SD advertisement MUST be accompanied by a TXT record of name 'api\_auth' with a value of either 'true' or 'false' dependent on whether authorization is needed for Node API requests.
```

docs/3.2. Discovery - Peer to Peer Operation.md
```diff
@@ -23,23 +22,23 @@ The IP address and port of the Node API MUST be identified via the DNS-SD advert
 
 Multiple DNS-SD advertisements for the same API are permitted where the API is exposed via multiple ports and/or protocols.
 
-Node mDNS announcements are discouraged from v1.3+ except for when using peer to peer mode as they generate unnecessary network traffic. Where a Node supports multiple versions including v1.2 or below it is still required to produce this mDNS advertisement in order to comply with earlier specification versions.
+From v1.3, Node mDNS announcements SHOULD only be used for peer to peer mode as they generate unnecessary network traffic. However a Node supporting multiple versions including v1.2 or below MUST still provide mDNS announcements as they were compulsory prior to v1.3.
 
 ## DNS-SD TXT records
 
-**api\_proto**
+### api\_proto
```
garethsb: This is the right level of heading indent here, but see question of heading or bold above.

docs/4.1. Behaviour - Registration.md
```diff
@@ -10,106 +10,107 @@ This document covers expected client and server behaviour for interactions betwe
...
 
 #### Resilient Node Behaviour
 
-In some deployments, it may be desirable to send multiple copies of essence (video/audio/data) via redundant paths in order to provide guarantees about availability. In these situations, it is likely that a single Node will be used at the point of content creation and transmission via the two paths, and again a single Node used for reception from the two paths.
+In some deployments, multiple copies of essence (video/audio/data) are sent via redundant paths in order to provide guarantees about availability. In these situations, it is likely that a single Node will be used at the point of content creation and transmission via the two paths, and again a single Node used for reception from the two paths.
 
-The data model employed by NMOS allows for this situation by modelling a single Flow sent via two independent Senders which are assigned to separate physical network interfaces on a Node.
+NMOS models this as a single Flow sent via two independent Senders which are assigned to separate physical network interfaces on a Node.
```
garethsb: This has led to confusion for some readers, since although the IS-04 model _allows_ this representation, ST 2022-7 is actually modelled by a _single_ Sender having two `interface_bindings`.

docs/4.1. Behaviour - Registration.md
```diff
@@ -10,106 +10,107 @@ This document covers expected client and server behaviour for interactions betwe
...
 
-This mode of operation is applicable to both the 'Peer to Peer' and 'Registered' discovery case. It is also envisaged in the Registered case that operation may be desirable with a single registry, or with two which operate totally independently on the two different networks.
+This mode of operation is applicable to both 'peer-to-peer' and 'registered' discovery. Registered discovery MAY use either a single registry, or a independent registry for each network.
```
garethsb:
```suggestion
This mode of operation is applicable to both 'peer-to-peer' and 'registered' discovery. Registered discovery MAY use either a single registry, or an independent registry for each network.
```

docs/4.1. Behaviour - Registration.md
```diff
@@ -10,106 +10,107 @@ This document covers expected client and server behaviour for interactions betwe
...
+In order to support this mode and ensure persistence in the registry upon link failure, a Node operating in this mode MAY register with Registration APIs (and host its Node API) via multiple independent network interfaces. If separate registries are deployed for the multiple network links the Node MUST ensure it registers the correct 'href' with each registry (see [Node API](../APIs/NodeAPI.raml) `/self` resource). From v1.1, the 'api' attribute of the Node /self resource adds the capability to list multiple endpoints for the same Node API, removing the requirement for this 'href' specialisation.
```
garethsb: I believe the business with registering the correct 'href' is wrong, as it breaks the identity rules. Let's discuss with @andrewbonney, next week.

docs/4.2. Behaviour - Querying.md
```diff
@@ -2,63 +2,63 @@
...
 
-There is no mandated URL base path for servers to use when creating websocket connections. Instead clients should observe the value of 'ws_href' which is returned by their subscription request in order to identify what to connect to.
+There is no mandated URL base path for servers to use when creating WebSocket connections. Instead clients SHOULD observe the value of 'ws_href' which is returned by their subscription request in order to identify what to connect to.
 
-Websockets created for specific clients will be cleaned up automatically if they disconnect.
+WebSockets created for specific clients will be cleaned up automatically if they disconnect.
 
-Connection upgrade procedures permitting an existing HTTP query to transition to a websocket are not currently supported, but may be in the future.
+Connection upgrade procedures permitting an existing HTTP query to transition to a WebSocket are not currently supported, but could be in the future.
```
garethsb: Maybe change "supported" to indicate "covered by this specification" better? Although, see next section, first para!

docs/4.2. Behaviour - Querying.md
```diff
@@ -2,63 +2,63 @@
...
 
 ### Query API Behaviour
-Long-lived connections to the Query API are supported via Websockets which can be set up via the /subscriptions resource. Query APIs may additionally support the HTTP 'Upgrade' header sent by clients to upgrade an HTTP GET request to a Websocket. In cases where this is performed, a corresponding entry in /subscriptions must also be created with matching query parameters.
 
-Upon receiving a request for a new subscription, the Query API should identify whether it already has any websockets open which provide for the requested query. If a websocket exists, the 'subscription' object representing it may be returned to the user. If a relevant websocket does not exist a new one should be created.
+Long-lived connections to the Query API are supported via WebSockets which can be set up via the /subscriptions resource. Query APIs MAY additionally support the HTTP 'Upgrade' header sent by clients to upgrade an HTTP GET request to a WebSocket. In cases where this is performed, a corresponding entry in /subscriptions MUST also be created with matching query parameters.
```
garethsb: Hmm, I thought we'd dealt with the fact that the sentence about HTTP 'Upgrade' here seems to cover precisely the thing that the previous section isn't currently supported... Need to check previous discussion.

docs/4.2. Behaviour - Querying.md
```diff
@@ -2,63 +2,63 @@
...
 
-In normal operation, once a websocket has no more clients subscribed to it the Query API may automatically remove it and its corresponding HTTP-advertised subscription object. The Query API MUST NOT acknowledge HTTP DELETE requests for websockets running in this 'non-persistent' mode, instead issuing an HTTP 403 response.
+Upon receiving a request for a new subscription, the Query API SHOULD identify whether it already has any WebSockets open which provide for the requested query. If a WebSocket exists, the 'subscription' object representing it MAY be returned to the user. If a relevant WebSocket does not exist a new one SHOULD be created.
```
garethsb: What does having a WebSocket "open" mean? Not a connection from a client, just whether or not it's currently listening for connections. Would be clearer if it just talked about having an appropriate subscription object...

docs/4.2. Behaviour - Querying.md
```diff
@@ -2,63 +2,63 @@
...
 
-If a 'persistent' websocket has been requested by the client, this must not be cleaned up automatically by the Query API, even if all consuming clients have disconnected. The user may request closing of this socket by issuing an HTTP DELETE. If an HTTP DELETE is issued prior to all websocket connections being closed, they should be forcibly closed by the server.
+In normal operation, once a WebSocket has no more clients subscribed to it the Query API MAY automatically remove it and its corresponding HTTP-advertised subscription object. The Query API MUST NOT acknowledge HTTP DELETE requests for WebSockets running in this 'non-persistent' mode, instead issuing an HTTP 403 response.
 
-### Websocket Protocol
+If a 'persistent' WebSocket has been requested by the client, this MUST NOT be cleaned up automatically by the Query API, even if all consuming clients have disconnected. The user MAY request closing of this socket by issuing an HTTP DELETE. If an HTTP DELETE is issued prior to all WebSocket connections being closed, they SHOULD be forcibly closed by the server.
```
garethsb: I have the same slight confusion here about what "a WebSocket" is. I'd rewrite those paragraphs with WebSocket replaced by subscription (and e.g. change 'running' to 'configured').

garethsb (commented):
Part 2 review. **docs/5.1. Data Model - Identifier Mapping.md** next.

docs/5.1. Data Model - Identifier Mapping.md
```diff
@@ -20,30 +20,30 @@ All resources shown above are part of the JT-NM reference architecture. Each log
...
 
-**Important Note:** A processing element which has a 1:1 input to output relationship and doesn't make editorial changes only need create a new Flow ID on output and not a new Source ID. As a Flow ID must descend from a single consistent Source ID, the Flow ID for these processing elements MUST be generated based on seed data including the incoming Flow ID. If this dynamic Flow generation cannot be achieved, the processing element MUST be viewed as a new Source instead (see Flow ID section below).
+**Important Note:** A processing element which has a 1:1 input to output relationship and doesn't make editorial changes only need create a new Flow ID on output and not a new Source ID. As a Flow ID descends from a single consistent Source ID, the Flow ID for these processing elements MUST be generated based on seed data including the incoming Flow ID. If this dynamic Flow generation cannot be achieved, the processing element MUST be viewed as a new Source instead (see Flow ID section below).
```
garethsb: Discuss next week. I don't think it's valid for a Node to have a Flow with a `source_id` for a Source on a different Node is it?

docs/5.1. Data Model - Identifier Mapping.md
```diff
@@ -53,28 +53,28 @@ UUIDs are used throughout the NMOS data model, and can be generated in a number
...
+-   If re-deploying via the same controller which created it and an instantiation of this virtual Node exists, it MUST be deployed with a new ID. Uses of this Node ID (and child resources) in any other stored data MUST be mapped to the new ID during this process.
+-   If re-deploying via a different controller and it cannot be guaranteed that another instance of this virtual Node does not already exist, a new Node ID MUST be generated, with any stored data being mapped to the new ID during the process.
 
-In order to simplify the above process for deploying snapshots, you may opt to generate a new Node ID for each instantiation of a virtual Node snapshot, but a mapping service must provide translations between old and new IDs as this may be required to keep session / production / automation data relevant.
+To simplify the above for deploying snapshots, each instantiation of a virtual Node snapshop MAY have a new Node ID. In such a case, a mapping MUST be provided to translate between old and new IDs so that session / production / automation data remain relevant.
```
garethsb: Saying that a mapping MUST be provided but not saying how that could be done in an open way is... not very helpful

docs/6.0. Upgrade Path.md
```diff
@@ -52,30 +52,30 @@ Nodes which support multiple versions simultaneously MUST ensure that all of the
 
 ## Requirements for Query API Clients
 
-Implementers of Query API clients MUST support at least one API version. Query API clients are strongly recommended to make use of the query downgrade mechanism in order to provide access to the widest possible range of Nodes' resources.
+Query API clients MUST support at least one API version. It is strongly RECOMMENDED that clients use of the query downgrade mechanism in order to provide access to the widest possible range of Nodes' resources.
```
garethsb:
```suggestion
Query API clients MUST support at least one API version. It is strongly RECOMMENDED that clients use the query downgrade mechanism in order to provide access to the widest possible range of Nodes' resources.
```

docs/6.0. Upgrade Path.md
```diff
@@ -52,30 +52,30 @@ Nodes which support multiple versions simultaneously MUST ensure that all of the
...
 
-Implementers should be aware that when operating a client against a Query API version lower than the maximum supported version by that Query API instance, they may be exposed to resource data beyond the scope of that API version's associated schemas. Clients SHOULD NOT attempt to access attributes which were only defined in a later specification version. Whilst the [Version Translation](#version-translations) mechanism should remove access to invalid additional keys, new values for existing keys are not excluded by this mechanism and should be handled gracefully.
+Clients SHOULD NOT attempt to access attributes introduced later Query API version. Whilst the [Version Translation](#version-translations) mechanism removes access to invalid additional keys, new values for existing keys are not excluded by this mechanism and need to be handled gracefully.
```
garethsb:
```suggestion
Clients SHOULD NOT attempt to access attributes introduced by a later Query API version. Whilst the [Version Translation](#version-translations) mechanism removes access to invalid additional keys, new values for existing keys are not excluded by this mechanism and need to be handled gracefully.
```

docs/2.0. APIs.md
peterbrightwell: Yes. I'll be proposing similar changes for IS-05 revision.

docs/3.2. Discovery - Peer to Peer Operation.md
peterbrightwell: Oops yes

docs/4.2. Behaviour - Querying.md
peterbrightwell: Will look at these. Agree that talking about subscriptions is more appropriate.

garethsb (commented):
Just a few more comments. That's it for a first pass.

docs/2.5. APIs - Query Parameters.md
peterbrightwell: On reflection this can go back to being a suggestion.

docs/2.5. APIs - Query Parameters.md
peterbrightwell: Well at least at the moment the linter doesn't check this sort of thing so lets leave it as pseudo-json for now

docs/2.5. APIs - Query Parameters.md
peterbrightwell: Agreed. i thought that someone would say this when i wrote it.

docs/2.5. APIs - Query Parameters.md
peterbrightwell: It's not really about granting permission so "can" better.

docs/2.5. APIs - Query Parameters.md
```diff
@@ -81,31 +86,35 @@ Payload Resources
...
-* The X-Paging headers identify properties of the collection of data returned in the response. These parameters may be used to construct a URL which would return the same set of bounded data on consecutive requests (for example /x-nmos/query/v1.1/nodes?paging.since=0:10&paging.until=0:20).
-* The Link header identifies the 'next' and 'prev' cursors which an application may use to make its next requests. An implementation may also provide 'first' and 'last' cursors to identify the beginning and end of a collection if this can be addressed via a consistent URL.
+
+- The data set returned when no query parameters are specified MUST be from the most recently updated (or created) resources in the collection, returned in descending order.
+- The X-Paging headers identify properties of the collection of data returned in the response. These parameters MAY be used to construct a URL which would return the same set of bounded data on consecutive requests (for example /x-nmos/query/v1.1/nodes?paging.since=0:10&paging.until=0:20).
```
peterbrightwell:
```suggestion
- The X-Paging headers identify properties of the collection of data returned in the response. These parameters can be used to construct a URL which would return the same set of bounded data on consecutive requests (for example /x-nmos/query/v1.1/nodes?paging.since=0:10&paging.until=0:20).
```

docs/2.5. APIs - Query Parameters.md
peterbrightwell:
```suggestion
```json
```

docs/2.5. APIs - Query Parameters.md
peterbrightwell: ...ought to be..." a bit of a strange construction for non-native speakers. What about "...are expected to be...

peterbrightwell:
```suggestion
When a client requests data which falls at the extreme ends of the stored data set it can be less clear what values are expected to be returned in the 'X-Paging-Limit' and 'X-Paging-Since' headers. The following examples are intended to clarify these cases.
```

docs/2.5. APIs - Query Parameters.md
peterbrightwell: As the first sentence in this para states a fact we can do the same for the second sentence.

peterbrightwell:
```suggestion
Basic queries make use of standard HTTP GET query parameters in the form '?key1=value1&key2=value2'. Keys match any attribute which the API schemas indicate could be returned by a given resource.
```

docs/2.5. APIs - Query Parameters.md
peterbrightwell: That would be a good thing. As things stand there is opportunity for non-interoperability here!

docs/2.5. APIs - Query Parameters.md
peterbrightwell: Probably not,  and maybe we should move to back-ticks for this sort of thing anyway as we've done in more recent specs.

peterbrightwell:
```suggestion
- This process continues up to the generation limit returned in the X-Ancestry-Generations header (in this case limited to 2 by a query parameter). The server MAY refuse to honour a requested '?query.ancestry_generations' value using a 400 error code if it is deemed to be too resource intensive.
```

docs/4.1. Behaviour - Registration.md
peterbrightwell: True, but out of scope of this PR

docs/5.1. Data Model - Identifier Mapping.md
peterbrightwell: Agreed. Let's discuss next week

docs/1.0. Overview.md
```diff
@@ -29,9 +42,9 @@ Clients requiring data about other Nodes in the system (such as connection manag
 
 Regardless of their implementation, viewed logically, Nodes provide:
 
-* An HTTP API to allow clients to view and manipulate the Node data model.
-* Interfaces (in the logical sense) through which content is transported.
-* A PTP slave for timing and synchronization.
+- An HTTP API to allow clients to view and manipulate the Node data model.
+- Interfaces (in the logical sense) through which content is transported.
+- A PTP slave for timing and synchronization.
```
garethsb: Did this ought to be "follower"?

peterbrightwell: Yes, and a later commit fixes this and also a description in clock_ptp.json

peterbrightwell, line 47:
```suggestion
- A PTP follower for timing and synchronization.
```

docs/4.1. Behaviour - Registration.md
garethsb:
```suggestion
NMOS can model this as a single Flow sent via two independent Senders which are assigned to separate physical network interfaces on a Node.
```

docs/4.1. Behaviour - Registration.md
garethsb: And if not noted in the backticks PR, there are more instances of "/self" and "'href'" in that sentence that could do with being quoted/backticked consistently.

docs/3.1. Discovery - Registered Operation.md
garethsb: Dealt with in the backticks child PR

docs/3.1. Discovery - Registered Operation.md
garethsb: Resolved on the backticks child PR.

docs/2.5. APIs - Query Parameters.md
garethsb: Agreed.

docs/2.5. APIs - Query Parameters.md
garethsb: Agreed

docs/2.5. APIs - Query Parameters.md
garethsb: Resolved in backticks child PR.

docs/2.5. APIs - Query Parameters.md
garethsb:
```suggestion
The choice to page based on 'creation' or 'update' timestamp depends on the client's intended use of the data:
```

docs/2.5. APIs - Query Parameters.md
garethsb: This conversation was marked as resolved, but I don't see an actual commit or suggestion to change 'MAY' to 'can'? Let's discuss this one tomorrow.

docs/2.5. APIs - Query Parameters.md
garethsb: Done

docs/4.1. Behaviour - Registration.md
garethsb: Done

docs/2.5. APIs - Query Parameters.md
garethsb: Done

docs/2.5. APIs - Query Parameters.md
garethsb: Done

garethsb:
Things that I think are outstanding:

* [x] Writing up an NMOS specifications style guide
* [x] Copying changes to common language in the early sections to other specs that repeat it (I've checked against IS-05 docs-language PR, but not other specs)
* [x] Rewriting the stuff about Query API Subscriptions/WebSocket connections
* [x] Maybe adding a note about `first` and `last` - https://github.com/AMWA-TV/nmos-discovery-registration/pull/158#discussion_r605825733
* [x] Thinking about a BCP for RQL - https://github.com/AMWA-TV/nmos-discovery-registration/pull/158#discussion_r605836088
* [x] Looking at the text around `href` specialisation - https://github.com/AMWA-TV/nmos-discovery-registration/pull/158#discussion_r606086882
* [x] Connection `Upgrade` for converting to WebSocket... in or out? - https://github.com/AMWA-TV/nmos-discovery-registration/pull/158#discussion_r606086882
* [x] Flows for other Nodes' Sources? - https://github.com/AMWA-TV/nmos-discovery-registration/pull/158#discussion_r606143745
* [x] Virtual Node id mapping - https://github.com/AMWA-TV/nmos-discovery-registration/pull/158#discussion_r606144352

docs/4.2. Behaviour - Querying.md
garethsb: It was #149. Done.

docs/2.5. APIs - Query Parameters.md
garethsb: Done

docs/2.5. APIs - Query Parameters.md
garethsb: Done

docs/4.1. Behaviour - Registration.md
andrewbonney:
```suggestion
In order to support this mode and ensure persistence in the registry upon link failure, a Node operating in this mode MAY register with Registration APIs (and host its Node API) via multiple independent network interfaces.

Clients are encouraged to use the `api` attribute rather than the deprecated `href` attribute when accessing a Node API (see [Node API](../APIs/NodeAPI.raml) `/self` resource) as the `href` attribute is likely to target a single network interface only. Nodes which need to support this mode of operation at v1.0 MUST register a different `href` for the same Node `id` with each registry, even though this is a workaround which technically breaches data model constraints.
```

See what you think of the above. I've tried to persist the normative language as-is, just changing the emphasis to new implementations rather than the required v1.0 behaviour.

garethsb: I tweaked a few single quotes to backticks.

I think the reemphasis is good.

W.r.t. "is likely to target a single network interface only", I was wondering how it couldn't... Using a DNS name, I suppose?
Maybe rather than "noting that" (who is noting?), something like "even though"?
I was looking again for the statement about `id` plus `version` standing for a unique variant of a resource... It's in the Common Keys section.

andrewbonney: Yeah, the DNS name was in my head there, but I suspect in practice implementations may use the first result found via DNS rather than iterating through them on failure, hence not calling that out as an explicit recommendation.

peterbrightwell: This is good.  I agree that "even though" would be a better way to phrase "noting that".

docs/5.1. Data Model - Identifier Mapping.md
```diff
@@ -53,139 +53,136 @@ UUIDs are used throughout the NMOS data model, and can be generated in a number
...
+- If re-deploying via the same controller which created it and an instantiation of this virtual Node exists, it MUST be deployed with a new ID. Uses of this Node ID (and child resources) in any other stored data MUST be mapped to the new ID during this process.
+- If re-deploying via a different controller and it cannot be guaranteed that another instance of this virtual Node does not already exist, a new Node ID MUST be generated, with any stored data being mapped to the new ID during the process.
 
-In order to simplify the above process for deploying snapshots, you may opt to generate a new Node ID for each instantiation of a virtual Node snapshot, but a mapping service must provide translations between old and new IDs as this may be required to keep session / production / automation data relevant.
+To simplify the above for deploying snapshots, each instantiation of a virtual Node snapshop MAY have a new Node ID. In such a case, a mapping MUST be provided to translate between old and new IDs so that session / production / automation data remain relevant.
```
peterbrightwell: Avoids using "must" or similar:

```suggestion
To simplify the above for deploying snapshots, each instantiation of a virtual Node snapshot MAY have a new Node ID. In such cases implementers are encouraged to consider how to translate between old and new IDs where session, production or automation data need to remain relevant.
```

garethsb: Also s/snapshop/snapshot/

garethsb: "each instantiation of a virtual Node snapshot MAY have a new Node ID" taken alone is a little confusing because _concurrent_ instantiations _MUST_ have different Node IDs.

docs/4.1. Behaviour - Registration.md
```diff
@@ -10,106 +10,107 @@ This document covers expected client and server behaviour for interactions betwe
...
+In order to support this mode and ensure persistence in the registry upon link failure, a Node operating in this mode MAY register with Registration APIs (and host its Node API) via multiple independent network interfaces. If separate registries are deployed for the multiple network links the Node MUST ensure it registers the correct `href` with each registry (see [Node API](../APIs/NodeAPI.raml) `/self` resource). From v1.1, the `api` attribute of the Node `/self` resource adds the capability to list multiple endpoints for the same Node API, removing the requirement for this `href` specialisation.
```
garethsb: Andrew's suggestion can't be applied because GitHub says it's outdated, so here it is again:

```suggestion
In order to support this mode and ensure persistence in the registry upon link failure, a Node operating in this mode MAY register with Registration APIs (and host its Node API) via multiple independent network interfaces.

Clients are encouraged to use the `api` attribute rather than the deprecated `href` attribute when accessing a Node API (see [Node API](../APIs/NodeAPI.raml) `/self` resource) as the `href` attribute is likely to target a single network interface only. Nodes which need to support this mode of operation at v1.0 MUST register a different `href` for the same Node `id` with each registry, even though this is a workaround which technically breaches data model constraints.
```

docs/4.1. Behaviour - Registration.md
garethsb: Done.

docs/5.1. Data Model - Identifier Mapping.md
peterbrightwell: Yes. Perhaps we should just simplify lines 58-64 significantly to e.g.:

For virtual Nodes, the Node ID MUST be universally unique to that Node. In cases where it is necessary to redeploy a snapshot of a virtual Node, a new Node ID might be required, together with a mechanism to map between the old and new IDs.

peterbrightwell: Obviously that omits previous "requirements", but it's not clear that they were actually mandated/followed.

docs/5.1. Data Model - Identifier Mapping.md
peterbrightwell: Resolving as this continued on a later conversation

docs/5.1. Data Model - Identifier Mapping.md
garethsb: Makes sense to me

peterbrightwell: Ok, I've done that in 500c75d6a5ed73f4b7aa453b541e727e51d9483a. I'll leave this conversation open for a bit so @andrewbonney sees it.

docs/3.2. Discovery - Peer to Peer Operation.md
```diff
@@ -2,78 +2,79 @@
...
+This document describes usage of NMOS APIs for discovery in cases where where a distributed registry is not available, such as small ad-hoc installations. This feature is OPTIONAL from v1.3 of the specification.
 
-Note that the peer-to-peer discovery mechanism depends upon mDNS which is not intended to operate over IP routed boundaries (ie. between network subnets). As such it is recommended that for most use cases, in particular fixed installations and those requiring high levels of resilience, the registered discovery mechanism is used.
+Note that the peer-to-peer discovery mechanism depends upon mDNS which is not intended to operate over IP routed boundaries (i.e. between network subnets). As such it is RECOMMENDED that for most use cases, in particular fixed installations and those requiring high levels of resilience, the registered discovery mechanism is used.
 
 ## Pre-Requisites
 
-The following pre-requisites are optional aspects of the [Node API](../APIs/NodeAPI.raml) implementation which must be in place to permit peer-to-peer discovery.
+When configured for peer-to-peer discovery, a Node MUST advertise its presence wth mDNS.
```
andrewbonney:
```suggestion
When configured for peer-to-peer discovery, a Node MUST advertise its presence with mDNS.
```

docs/5.1. Data Model - Identifier Mapping.md
```diff
@@ -4,46 +4,46 @@ _(c) AMWA 2016, CC Attribution-NoDerivatives 4.0 International (CC BY-ND 4.0)_
...
 
-**Important Note:** A processing element which has a 1:1 input to output relationship and doesn't make editorial changes only need create a new Flow ID on output and not a new Source ID. As a Flow ID must descend from a single consistent Source ID, the Flow ID for these processing elements MUST be generated based on seed data including the incoming Flow ID. If this dynamic Flow generation cannot be achieved, the processing element MUST be viewed as a new Source instead (see Flow ID section below).
+**Important Note:** A processing element which has a 1:1 input to output relationship and doesn't make editorial changes only need create a new Flow ID on output and not a new Source ID. As a Flow ID descends from a single consistent Source ID, the Flow ID for these processing elements MUST be generated based on seed data including the incoming Flow ID. If this dynamic Flow generation cannot be achieved, the processing element MUST be viewed as a new Source instead (see Flow ID section below).
```
peterbrightwell: [As per discussion] `source_id` could match a different Node's Source ID, but the above still stands. It could be written more clearly though:

```suggestion
Where there is a one-to-one input-output relationship with no editorial change, the output only needs to have a new Flow ID, as the same Source ID can be used. In this case the new Flow ID MUST be generated from information that includes the incoming Flow ID. If this cannot be implemented then a new Source ID MUST be used instead. For more details see the Flow ID section below.
```

@andrewbonney @garethsb I realise this omits the "descends from" part but does that really help?

andrewbonney: Looks like there may be a missing word/words in "In this case the new Flow ID MUST be generated from information that the incoming Flow ID"

peterbrightwell: Yes fixed so it includes "includes".

andrewbonney: LGTM

docs/2.1. APIs - Common Keys.md
peterbrightwell: Made #161 to discuss.

docs/4.2. Behaviour - Querying.md
peterbrightwell:
```suggestion
Upon receiving a request for a new subscription, the Query API SHOULD identify whether it already has a `subscription` object which provides for the requested query, and if so MAY return the `subscription` object to the user. If a relevant `subscription` object does not exist a new one SHOULD be created.
```

docs/4.2. Behaviour - Querying.md
```diff
@@ -2,63 +2,63 @@
...
+Upon receiving a request for a new subscription, the Query API SHOULD identify whether it already has any WebSockets open which provide for the requested query. If a WebSocket exists, the 'subscription' object representing it MAY be returned to the user. If a relevant WebSocket does not exist a new one SHOULD be created.
 
-If a 'persistent' websocket has been requested by the client, this must not be cleaned up automatically by the Query API, even if all consuming clients have disconnected. The user may request closing of this socket by issuing an HTTP DELETE. If an HTTP DELETE is issued prior to all websocket connections being closed, they should be forcibly closed by the server.
+In normal operation, once a WebSocket has no more clients subscribed to it the Query API MAY automatically remove it and its corresponding HTTP-advertised subscription object. The Query API MUST NOT acknowledge HTTP DELETE requests for WebSockets running in this 'non-persistent' mode, instead issuing an HTTP 403 response.
```
peterbrightwell:
```suggestion
The Query API MAY remove any `subscription` objects that no longer have clients. The Query API MUST NOT acknowledge HTTP `DELETE` requests for WebSockets running in this 'non-persistent' mode, instead issuing an HTTP `403` (Forbidden) response.
```

docs/4.2. Behaviour - Querying.md
peterbrightwell: Added suggested versions for two paras, but I'm not sure about the third one. I think this comes down to what 'persistent' can apply to. @andrewbonney any comments?

docs/2.5. APIs - Query Parameters.md
peterbrightwell: See also #20

peterbrightwell:
https://github.com/AMWA-TV/nmos-template/issues/19 suggests a style guide

docs/4.2. Behaviour - Querying.md
peterbrightwell: Would be better resolved through #162

docs/4.2. Behaviour - Querying.md
peterbrightwell: Would be better resolved through #162

peterbrightwell: Would be better resolved through #162

docs/4.2. Behaviour - Querying.md
peterbrightwell: Would be better resolved through #162

docs/2.5. APIs - Query Parameters.md
```diff
@@ -308,248 +344,291 @@ Payload Resources
...
+## Advanced (RQL) Queries (OPTIONAL)
 
-RQL should be formatted in the normalised form as opposed to using FIQL syntax, and passed via the query string using '?query.rql=...'.
+Query APIs MAY support Resource Query Language (RQL, <https://github.com/persvr/rql/>) queries against their API resources where the `rql` trait is specified in the RAML documentation. A `501` (Not Implemented) HTTP status code MUST be returned where an RQL query is attempted using RQL functions or operators which are not supported by a Query API.
 
-Querying within objects and arrays should be performed as in the Basic Queries case by using the '.' separator.
+RQL SHOULD be formatted in the normalised form as opposed to using FIQL syntax, and passed via the query string using a `query.rql=...` query parameter.
+
+Querying within objects and arrays SHOULD be performed as in the Basic Queries case by using the `.` separator.
```
garethsb:
```suggestion
Querying within objects and arrays is performed as in the Basic Queries case by using the `.` separator.
```

docs/3.1. Discovery - Registered Operation.md
```diff
@@ -8,86 +8,98 @@ This document describes usage of NMOS APIs for discovery in cases where where a
 
 ### DNS-SD Advertisement
 
-The preferred method of Registration API advertisement is via unicast DNS-SD advertisement of the type \_nmos-register.\_tcp. Registration APIs MUST additionally be capable of producing an mDNS advertisement. This MAY be disabled via a user-configurable method.
+The preferred method of Registration API advertisement is via unicast DNS-SD advertisement of the type `_nmos-register._tcp`. Registration APIs MUST additionally be capable of producing an mDNS advertisement. This MAY be disabled via a user-configurable method.
 
-Registration APIs supporting versions 'v1.2' and below MUST additionally be capable of producing an mDNS advertisement of type \_nmos-registration.\_tcp. This MAY be disabled via a user-configurable method.
-*note: RFC6763 Section 7.2 specifies that the maximum service name length for an mDNS advertisement is 16 characters when including the leading underscore, but "_nmos-registration" is 18 characters.*
+Registration APIs supporting versions `v1.2` and below MUST additionally be capable of producing an mDNS advertisement of type `_nmos-registration._tcp`. This MAY be disabled via a user-configurable method.
```
garethsb: Does the MAY here apply to the server or the client/user? I.e. Is it valid for a server not to support this user configurability? 

Actually that question applies to several other places that say "This MAY be disabled via a user-configurable method."

garethsb (approved):
Done. 🤞

PR #159, opened 2021-04-05
Use backticks to render URLs etc as monospaced

peterbrightwell:
Currently these use 'quotes' for URLs, parameter names etc, but more recent NMOS specs have used `backticks` for monospace rendering., This is consistent how examples and other codeblocks appear and look better.

This should be merged (if we choose to do it) before merging docs-language. Or is it `docs-language`?

### DIFF
--- a/docs/1.0. Overview.md
+++ b/docs/1.0. Overview.md
@@ -70,9 +70,9 @@ Registered discovery takes place using a **Registration & Discovery System (RDS)
 
 ![Registration and Discovery](images/registration-and-discovery.png)
 
-The Registration Service implements the Registration API of the NMOS Discovery and Registration Specification. Nodes POST to this API to register themselves and their resources. The Registration Service also manages garbage collection of Nodes and their resources by requiring Nodes to send regular keep-alive/heartbeat messages.
+The Registration Service implements the Registration API of the NMOS Discovery and Registration Specification. Nodes `POST` to this API to register themselves and their resources. The Registration Service also manages garbage collection of Nodes and their resources by requiring Nodes to send regular keep-alive/heartbeat messages.
 
-The Query Service implements the Query API of the NMOS Discovery and Registration Specification. Clients can GET lists of resources from this API. Typical usage examples include:
+The Query Service implements the Query API of the NMOS Discovery and Registration Specification. Clients can `GET` lists of resources from this API. Typical usage examples include:
 
 - Obtaining a list of registered Nodes in order to drive a configuration interface.
 - Obtaining a list of Sender resources and a list of Receiver resources in order to provide a connection management interface.
--- a/docs/2.0. APIs.md
+++ b/docs/2.0. APIs.md
@@ -27,7 +27,7 @@ All APIs MUST provide a JSON representation signalled via `Content-Type: applica
 
 ### API Paths
 
-All NMOS APIs MUST use a path in the following format. Other HTTP resources MAY be presented on the same port by the Node, provided all NMOS resources are available from the /x-nmos/ path as follows:
+All NMOS APIs MUST use a path in the following format. Other HTTP resources MAY be presented on the same port by the Node, provided all NMOS resources are available from the `/x-nmos/` path as follows:
 
 ```
 http(s)://<ip address or hostname>:<port>/x-nmos/<api type>/<api version>/
--- a/docs/2.0. APIs.md
+++ b/docs/2.0. APIs.md
@@ -111,6 +111,6 @@ For HTTP codes 400 and upwards, a JSON format response MUST be returned as follo
 }
 ```
 
-`code` SHOULD match the HTTP status code. `error` MUST always be present and in string format. `debug` MAY be null if no further debug information is available.
+`code` SHOULD match the HTTP status code. `error` MUST always be present and in string format. `debug` MAY be `null` if no further debug information is available.
 
 Further details on when APIs will respond with particular codes is covered in the [Behaviour](4.0.%20Behaviour.md) section.
--- a/docs/2.0. APIs.md
+++ b/docs/2.0. APIs.md
@@ -35,9 +35,9 @@ http(s)://<ip address or hostname>:<port>/x-nmos/<api type>/<api version>/
 
 At each level of the API from the base resource upwards, the response SHOULD include a JSON format array of the child resources available from that point.
 
-API clients using DNS-SD for API discovery SHOULD construct these paths by making use of the TXT records 'api\_proto' and 'api\_ver', along with addresses and ports resolved via DNS-SD. API clients which are discovering Node APIs via a Query API SHOULD construct Node API paths using the corresponding data available within the `api` attributes within the Query API `/nodes/` path.
+API clients using DNS-SD for API discovery SHOULD construct these paths by making use of the TXT records `api_proto` and `api_ver`, along with addresses and ports resolved via DNS-SD. API clients which are discovering Node APIs via a Query API SHOULD construct Node API paths using the corresponding data available within the `api` attributes within the Query API `/nodes/` path.
 
-Where protocol and versioning data is not available, clients MAY assume a v1.0 API, which operates via the `http` protocol only.
+Where protocol and versioning data is not available, clients MAY assume a v1.0 API, which operates via HTTP only.
 
 ### Versioning
 
@@ -69,7 +69,7 @@ A v1.1 API response can include:
 - The versioning format is `v<#MAJOR>.<#MINOR>`
 - MINOR increments will be performed for non-breaking changes (such as the addition of attributes in a response)
 - MAJOR increments will be performed for breaking changes (such as the renaming of a resource or attribute)
-- Versions MUST be represented as complete strings. Parsing MUST proceed as follows: separate into two strings, using the point (.) as a delimiter. Compare integer representations of MAJOR, MINOR version (such that v1.12 is greater than v1.5).
+- Versions MUST be represented as complete strings. Parsing MUST proceed as follows: separate into two strings, using the point (`.`) as a delimiter. Compare integer representations of MAJOR, MINOR version (such that v1.12 is greater than v1.5).
 - Clients are responsible for identifying the correct API version they require.
 
 ### URLs: Approach to Trailing Slashes
--- a/docs/2.0. APIs.md
+++ b/docs/2.0. APIs.md
@@ -29,20 +29,21 @@ All APIs MUST provide a JSON representation signalled via 'Content-Type: applica
 
 All NMOS APIs MUST use a path in the following format. Other HTTP resources MAY be presented on the same port by the Node, provided all NMOS resources are available from the /x-nmos/ path as follows:
 
+```
 http(s)://<ip address or hostname>:<port>/x-nmos/<api type>/<api version>/
 ```
 
 At each level of the API from the base resource upwards, the response SHOULD include a JSON format array of the child resources available from that point.
 
-API clients using DNS-SD for API discovery SHOULD construct these paths by making use of the TXT records 'api\_proto' and 'api\_ver', along with addresses and ports resolved via DNS-SD. API clients which are discovering Node APIs via a Query API SHOULD construct Node API paths using the corresponding data available within the \'api\' attributes within the Query API /nodes/ path.
+API clients using DNS-SD for API discovery SHOULD construct these paths by making use of the TXT records 'api\_proto' and 'api\_ver', along with addresses and ports resolved via DNS-SD. API clients which are discovering Node APIs via a Query API SHOULD construct Node API paths using the corresponding data available within the `api` attributes within the Query API `/nodes/` path.
 
-Where protocol and versioning data is not available, clients MAY assume a v1.0 API, which operates via the 'http' protocol only.
+Where protocol and versioning data is not available, clients MAY assume a v1.0 API, which operates via the `http` protocol only.
 
 ### Versioning
 
 All public APIs are versioned as follows:
 
-- Requesting the API base resource (such as http(s)://&lt;ip address or hostname&gt;:&lt;port&gt;/x-nmos/query/) will provide a list containing the versions of the API present on the node.
+- Requesting the API base resource (such as `http(s)://<ip address or hostname>/x-nmos/query/`) will provide a list containing the versions of the API present on the node.
 - A versioned API response MUST include only resources which match the schema for that API version.
 - Data which is held for mismatched minor API versions MAY be returned if it can be conformed to the correct schema (see example below). Data MUST NOT be conformed between major API versions.
 
@@ -64,8 +65,8 @@ A v1.1 API response can include:
 ]
 ```
 
-- Appending /v1.0/ to the API base resource will request version 1.0 of the API if available.
-- The versioning format is v&lt;#MAJOR&gt;.&lt;#MINOR&gt;
+- Appending `/v1.0/` to the API base resource will request version 1.0 of the API if available.
+- The versioning format is `v<#MAJOR>.<#MINOR>`
 - MINOR increments will be performed for non-breaking changes (such as the addition of attributes in a response)
 - MAJOR increments will be performed for breaking changes (such as the renaming of a resource or attribute)
 - Versions MUST be represented as complete strings. Parsing MUST proceed as follows: separate into two strings, using the point (.) as a delimiter. Compare integer representations of MAJOR, MINOR version (such that v1.12 is greater than v1.5).
@@ -92,7 +93,7 @@ In order to overcome shortcomings in some common libraries, the following requir
 
 #### Client Behaviour
 
-When a server implementation needs to indicate an API URL via an 'href' or similar attribute, it is valid to either include a trailing slash or not provided that the listed path is accessible and follows the above rules. Clients appending paths to 'href' type attributes MUST support both cases, avoiding doubled or missing slashes.
+When a server implementation needs to indicate an API URL via an `href` or similar attribute, it is valid to either include a trailing slash or not provided that the listed path is accessible and follows the above rules. Clients appending paths to `href` type attributes MUST support both cases, avoiding doubled or missing slashes.
 
 ### Error Codes & Responses
 
@@ -110,6 +111,6 @@ For HTTP codes 400 and upwards, a JSON format response MUST be returned as follo
 }
 ```
 
-The 'code' SHOULD match the HTTP status code. 'error' MUST always be present and in string format. 'debug' MAY be null if no further debug information is available.
+`code` SHOULD match the HTTP status code. `error` MUST always be present and in string format. `debug` MAY be null if no further debug information is available.
 
 Further details on when APIs will respond with particular codes is covered in the [Behaviour](4.0.%20Behaviour.md) section.
--- a/docs/2.0. APIs.md
+++ b/docs/2.0. APIs.md
@@ -19,7 +19,7 @@ Examples of JSON format output are provided in the [examples](../examples/) fold
 
 ## API Validation
 
-JSON schemas are included with the RAML API definitions. These include validation for values used within the APIs. It is strongly RECOMMENDED that implementers of a Registration API use these JSON schemas as part of a validation stage when receiving registrations from Nodes. Invalid resources SHOULD cause a 400 HTTP error (Bad Request) to be returned to the client.
+JSON schemas are included with the RAML API definitions. These include validation for values used within the APIs. It is strongly RECOMMENDED that implementers of a Registration API use these JSON schemas as part of a validation stage when receiving registrations from Nodes. Invalid resources SHOULD cause a 400 (Bad Request) HTTP error to be returned to the client.
 
 ### Content Types
 
--- a/docs/2.0. APIs.md
+++ b/docs/2.0. APIs.md
@@ -23,7 +23,7 @@ JSON schemas are included with the RAML API definitions. These include validatio
 
 ### Content Types
 
-All APIs MUST provide a JSON representation signalled via 'Content-Type: application/json' headers. This SHOULD be the default content type in the absence of any requested alternative by clients. Other content types (such as HTML) are permitted if they are explicitly requested via Accept headers.
+All APIs MUST provide a JSON representation signalled via `Content-Type: application/json` headers. This SHOULD be the default content type in the absence of any requested alternative by clients. Other content types (such as HTML) are permitted if they are explicitly requested via Accept headers.
 
 ### API Paths
 
--- a/docs/2.0. APIs.md
+++ b/docs/2.0. APIs.md
@@ -78,13 +78,13 @@ For consistency and in order to adhere to how these APIs are specified in RAML, 
 
 In order to overcome shortcomings in some common libraries, the following requirements are imposed for the support of URL paths with or without trailing slashes.
 
-#### GET and HEAD Requests
+#### `GET` and `HEAD` Requests
 
 - Clients performing requests using these methods SHOULD correctly handle a 301 response (moved permanently).
 - When a 301 is supported, the client MUST follow the redirect in order to retrieve the response payload.
 - Servers implementing the APIs MUST support requests using these methods to both the trailing slash and non-trailing slash path to each resource. One of these MAY produce a 301 response causing a redirect to the other if preferred.
 
-#### All Other Requests (PUT, POST, DELETE, OPTIONS etc)
+#### All Other Requests (`PUT`, `POST`, `DELETE`, `OPTIONS`, etc)
 
 - Clients performing requests using these methods MUST use URLs with no trailing slash present.
 - Servers implementing the APIs MUST correctly interpret requests using these methods to paths without trailing slashes present.
--- a/docs/2.0. APIs.md
+++ b/docs/2.0. APIs.md
@@ -35,7 +35,7 @@ http(s)://<ip address or hostname>:<port>/x-nmos/<api type>/<api version>/
 
 At each level of the API from the base resource upwards, the response SHOULD include a JSON format array of the child resources available from that point.
 
-API clients using DNS-SD for API discovery SHOULD construct these paths by making use of the TXT records `api_proto` and `api_ver`, along with addresses and ports resolved via DNS-SD. API clients which are discovering Node APIs via a Query API SHOULD construct Node API paths using the corresponding data available within the `api` attributes within the Query API `/nodes/` path.
+API clients using DNS-SD for API discovery SHOULD construct these paths by making use of the TXT records `api_proto` and `api_ver`, along with addresses and ports resolved via DNS-SD. API clients which are discovering Node APIs via a Query API SHOULD construct Node API paths using the corresponding data available within the `api` attributes within the Query API `/nodes` resources.
 
 Where protocol and versioning data is not available, clients MAY assume a v1.0 API, which operates via HTTP only.
 
--- a/docs/2.1. APIs - Common Keys.md
+++ b/docs/2.1. APIs - Common Keys.md
@@ -6,9 +6,9 @@ A number of common keys are used within NMOS APIs. Their definitions, semantics 
 
 ## Version
 
-Core resources such as Sources, Flows, Nodes etc. include a 'version' attribute. As properties of a given Flow or similar will change over its lifetime, the version identifies the instant at which this change took place. For example, if a vision mix transition occurred which resulted in a change to the 'parents' which make up a particular Flow, the 'version' is modified to reflect the wall clock instant at which this change occurred.
+Core resources such as Sources, Flows, Nodes etc. include a `version` attribute. As properties of a given Flow or similar will change over its lifetime, the version identifies the instant at which this change took place. For example, if a vision mix transition occurred which resulted in a change to the `parents` which make up a particular Flow, the `version` is modified to reflect the wall clock instant at which this change occurred.
 
-The version is a nanosecond TAI timestamp represented as a string of the format: &lt;seconds&gt;:&lt;nanoseconds&gt;. Note the ':' as opposed to a '.', indicating that the value '1439299836:10' is interpreted as 1439299836 seconds and 10 nanoseconds.
+`version` is a nanosecond TAI timestamp represented as a string of the format: `<seconds>:<nanoseconds`. Note the  `:` as opposed to a `.`, indicating that the value `1439299836:10` is interpreted as 1439299836 seconds and 10 nanoseconds.
 
 Combining the version with the resource's ID attribute creates a hash which can be used to uniquely identify a particular variant of a Source, Flow or similar.
 
@@ -24,12 +24,12 @@ Freeform human readable string to be used within user interfaces.
 
 A URN describing the data format of a video, audio or data flow.
 
-| **Permitted Value**     | **Valid From API Version** |
-|-------------------------|----------------------------|
-| urn:x-nmos:format:video | v1.0                       |
-| urn:x-nmos:format:audio | v1.0                       |
-| urn:x-nmos:format:data  | v1.0                       |
-| urn:x-nmos:format:mux   | v1.1                       |
+| **Permitted Value**       | **Valid From API Version** |
+|---------------------------|----------------------------|
+| `urn:x-nmos:format:video` | v1.0                       |
+| `urn:x-nmos:format:audio` | v1.0                       |
+| `urn:x-nmos:format:data`  | v1.0                       |
+| `urn:x-nmos:format:mux `  | v1.1                       |
 
 See [Use of URNs](#use-of-urns) below for additional requirements regarding their handling.
 
@@ -39,14 +39,14 @@ A URN describing the protocol used to send data (video, audio, data etc.) over a
 
 Note: From v1.3 onwards, permitted transport types are not versioned and are instead defined via the [NMOS Parameter Registers](https://github.com/AMWA-TV/nmos-parameter-registers). Prior to v1.3, the following table applies.
 
-| **Permitted Value**            | **Valid From API Version** |
-|--------------------------------|----------------------------|
-| urn:x-nmos:transport:rtp       | v1.0                       |
-| urn:x-nmos:transport:rtp.mcast | v1.0                       |
-| urn:x-nmos:transport:rtp.ucast | v1.0                       |
-| urn:x-nmos:transport:dash      | v1.0                       |
+| **Permitted Value**              | **Valid From API Version** |
+|----------------------------------|----------------------------|
+| `urn:x-nmos:transport:rtp`       | v1.0                       |
+| `urn:x-nmos:transport:rtp.mcast` | v1.0                       |
+| `urn:x-nmos:transport:rtp.ucast` | v1.0                       |
+| `urn:x-nmos:transport:dash`      | v1.0                       |
 
-For example, an RTP Transmitter sending to a multicast group uses the transport 'urn:x-nmos:transport:rtp.mcast', but a receiver supporting both unicast and multicast presents the transport 'urn:x-nmos:transport:rtp' to indicate its less restrictive state.
+For example, an RTP Transmitter sending to a multicast group uses the transport `urn:x-nmos:transport:rtp.mcast`, but a receiver supporting both unicast and multicast presents the transport `urn:x-nmos:transport:rtp` to indicate its less restrictive state.
 
 See [Use of URNs](#use-of-urns) below for additional requirements regarding their handling.
 
@@ -72,33 +72,33 @@ A URN describing the role which a Device has within the production environment (
 
 Note: From v1.3 onwards, permitted device types are not versioned and are instead defined via the [NMOS Parameter Registers](https://github.com/AMWA-TV/nmos-parameter-registers). Prior to v1.3, the following table applies.
 
-| **Permitted Value**        | **Valid From API Version** |
-|----------------------------|----------------------------|
-| urn:x-nmos:device:generic  | v1.0                       |
-| urn:x-nmos:device:pipeline | v1.0                       |
+| **Permitted Value**          | **Valid From API Version** |
+|------------------------------|----------------------------|
+| `urn:x-nmos:device:generic`  | v1.0                       |
+| `urn:x-nmos:device:pipeline` | v1.0                       |
 
 See [Use of URNs](#use-of-urns) below for additional requirements regarding their handling.
 
 ## Use of URNs
 
-All Uniform Resource Names (URNs) specified within the scope of the NMOS specifications MUST use the namespace 'urn:x-nmos:'. Manufacturers MAY use their own namespaces to indicate values of URN which are not currently defined within the NMOS namespace.
+All Uniform Resource Names (URNs) specified within the scope of the NMOS specifications MUST use the namespace `urn:x-nmos:`. Manufacturers MAY use their own namespaces to indicate values of URN which are not currently defined within the NMOS namespace.
 
-Where a URN begins 'urn:x-nmos:' it MUST use the following construction:
+Where a URN begins `urn:x-nmos:` it MUST use the following construction:
 
-- The URN 'base' is a sequence of one or more names delimited by ':' characters, including the namespace
+- The URN `base` is a sequence of one or more names delimited by `:` characters, including the namespace
 - The URN MAY include one or more 'subclassifications'
 - The URN MAY include a version number
 
-Subclassifications are defined as the portion of the URN which follows the first occurrence of a '.', but prior to any '/' character. By convention, URNs which begin 'urn:x-nmos:' will never include a further ':' after the first occurrence of a '.'.
+Subclassifications are defined as the portion of the URN which follows the first occurrence of a `.`, but prior to any `/` character. By convention, URNs which begin `urn:x-nmos:` will never include a further `:` after the first occurrence of a `.`.
 
-For example, the URN base 'urn:x-nmos:transport:rtp' above has two subclassifications of 'ucast' and 'mcast'.
+For example, the URN base `urn:x-nmos:transport:rtp` above has two subclassifications of `ucast` and `mcast`.
 
-API clients interpreting URNs MUST correctly recognise any as-yet undefined subclassifications as being associated with the URN base which appears prior to the first '.' character.
+API clients interpreting URNs MUST correctly recognise any as-yet undefined subclassifications as being associated with the URN base which appears prior to the first `.` character.
 
-For example, if filtering by the format 'urn:x-nmos:format:video', the filter includes any as-yet undefined subclassifications such as 'urn:x-nmos:format:video.raw'.
+For example, if filtering by the format `urn:x-nmos:format:video`, the filter includes any as-yet undefined subclassifications such as `urn:x-nmos:format:video.raw`.
 
-Versions are defined as the portion of the URN which follows the first occurrence of a '/', up to and including the remainder of the URN. Any '.' characters after this point indicate a version delimiter. By convention, URNs which begin 'urn:x-nmos:' will format the version as 'v<#MAJOR>.<#MINOR>' and SHOULD be handled as specified in [APIs](2.0.%20APIs.md).
+Versions are defined as the portion of the URN which follows the first occurrence of a `/`, up to and including the remainder of the URN. Any `.` characters after this point indicate a version delimiter. By convention, URNs which begin `urn:x-nmos:` will format the version as `v<#MAJOR>.<#MINOR>` and SHOULD be handled as specified in [APIs](2.0.%20APIs.md).
 
-For example, the URN 'urn:x-nmos:control:sr-ctrl/v1.0' can be separated into a base 'urn:x-nmos:control:sr-ctrl' and version 'v1.0'.
+For example, the URN `urn:x-nmos:control:sr-ctrl/v1.0` can be separated into a base `urn:x-nmos:control:sr-ctrl` and version `v1.0`.
 
 Query API clients MUST be tolerant to URNs which have not yet been defined, but which might be added in later API versions.
--- a/docs/2.1. APIs - Common Keys.md
+++ b/docs/2.1. APIs - Common Keys.md
@@ -85,7 +85,7 @@ All Uniform Resource Names (URNs) specified within the scope of the NMOS specifi
 
 Where a URN begins `urn:x-nmos:` it MUST use the following construction:
 
-- The URN `base` is a sequence of one or more names delimited by `:` characters, including the namespace
+- The URN base is a sequence of one or more names delimited by `:` characters, including the namespace
 - The URN MAY include one or more 'subclassifications'
 - The URN MAY include a version number
 
--- a/docs/2.1. APIs - Common Keys.md
+++ b/docs/2.1. APIs - Common Keys.md
@@ -29,7 +29,7 @@ A URN describing the data format of a video, audio or data flow.
 | `urn:x-nmos:format:video` | v1.0                       |
 | `urn:x-nmos:format:audio` | v1.0                       |
 | `urn:x-nmos:format:data`  | v1.0                       |
-| `urn:x-nmos:format:mux `  | v1.1                       |
+| `urn:x-nmos:format:mux`   | v1.1                       |
 
 See [Use of URNs](#use-of-urns) below for additional requirements regarding their handling.
 
--- a/docs/2.3. APIs - Server Side Implementation Notes.md
+++ b/docs/2.3. APIs - Server Side Implementation Notes.md
@@ -6,7 +6,7 @@ _(c) AMWA 2016, CC Attribution-NoDerivatives 4.0 International (CC BY-ND 4.0)_
 
 In order to permit web-based control interfaces to be hosted remotely, all NMOS APIs MUST implement valid CORS HTTP headers in responses to all requests.
 
-In addition to this, where highlighted in the API specifications servers MUST respond to HTTP pre-flight OPTIONS requests. Servers MAY additionally support HTTP OPTIONS requests made to any other API resource.
+In addition to this, where highlighted in the API specifications servers MUST respond to HTTP pre-flight `OPTIONS` requests. Servers MAY additionally support HTTP ``OPTIONS requests made to any other API resource.
 
 In order to simplify development, the following headers MAY be returned in order to remove these restrictions as far as possible. Note that these are very relaxed and possibly not suitable for a production deployment.
 
--- a/docs/2.3. APIs - Server Side Implementation Notes.md
+++ b/docs/2.3. APIs - Server Side Implementation Notes.md
@@ -17,12 +17,12 @@ Access-Control-Allow-Headers: Content-Type, Accept
 Access-Control-Max-Age: 3600
 ```
 
-To ensure compatibility, the response 'Access-Control-Allow-Headers' could be set from the request's 'Access-Control-Request-Headers'.
+To ensure compatibility, the response `Access-Control-Allow-Headers` could be set from the request's `Access-Control-Request-Headers`.
 
-## Use of '.local' Hostnames
+## Use of `.local` Hostnames
 
-Where 'href' or 'host' attributes are specified in the APIs, it is strongly RECOMMENDED to use either IP addresses or hostnames which are resolvable using unicast DNS. Whilst '.local' hostnames are convenient in a zero-configuration layer 2 network segment, these are not usually resolvable beyond these boundaries. As this specification is intended for use between layer 3 network segments, use of these hostnames can result in Nodes appearing inaccessible.
+Where `href` or `host` attributes are specified in the APIs, it is strongly RECOMMENDED to use either IP addresses or hostnames which are resolvable using unicast DNS. Whilst `.local` hostnames are convenient in a zero-configuration layer 2 network segment, these are not usually resolvable beyond these boundaries. As this specification is intended for use between layer 3 network segments, use of these hostnames can result in Nodes appearing inaccessible.
 
 ## Node API Senders: Use with RTP
 
-When using RTP for a sender, the 'manifest_href' MUST be an HTTP-accessible reference to an SDP file. As per [RFC 4566](https://tools.ietf.org/html/rfc4566), SDP files advertised via web services SHOULD use a MIME type of 'application/sdp'.
+When using RTP for a sender, the `manifest_href` MUST be an HTTP-accessible reference to an SDP file. As per [RFC 4566](https://tools.ietf.org/html/rfc4566), SDP files advertised via web services SHOULD use a MIME type of `application/sdp`.
--- a/docs/2.3. APIs - Server Side Implementation Notes.md
+++ b/docs/2.3. APIs - Server Side Implementation Notes.md
@@ -6,7 +6,7 @@ _(c) AMWA 2016, CC Attribution-NoDerivatives 4.0 International (CC BY-ND 4.0)_
 
 In order to permit web-based control interfaces to be hosted remotely, all NMOS APIs MUST implement valid CORS HTTP headers in responses to all requests.
 
-In addition to this, where highlighted in the API specifications servers MUST respond to HTTP pre-flight `OPTIONS` requests. Servers MAY additionally support HTTP ``OPTIONS requests made to any other API resource.
+In addition to this, where highlighted in the API specifications servers MUST respond to HTTP pre-flight `OPTIONS` requests. Servers MAY additionally support HTTP `OPTIONS` requests made to any other API resource.
 
 In order to simplify development, the following headers MAY be returned in order to remove these restrictions as far as possible. Note that these are very relaxed and possibly not suitable for a production deployment.
 
--- a/docs/2.5. APIs - Query Parameters.md
+++ b/docs/2.5. APIs - Query Parameters.md
@@ -249,7 +249,7 @@ Payload Resources
 
 When a client requests data which falls at the extreme ends of the stored data set it can be less clear what values SHOULD be returned in the `X-Paging-Limit` and `X-Paging-Since` headers. The following examples are intended to clarify these cases.
 
-Where a paging limit is not specified in a request the server`s default is used.
+Where a paging limit is not specified in a request the server's default is used.
 
 **Example 1: Client request occurs at the beginning of the data set**
 
--- a/docs/2.5. APIs - Query Parameters.md
+++ b/docs/2.5. APIs - Query Parameters.md
@@ -480,7 +480,7 @@ GET /x-nmos/query/v1.0/sources?format=urn:x-nmos:format:video&device_id=9126cc2f
 
 ***Response***
 
-- Returns all Source objects which have an attribute of `format` which exactly matches `urn:x-nmos:format:video` AND a `device_id` attribute which exactly matches `9126cc2f-4c26-4c9b-a6cd-93c4381c9be5`.
+- Returns all Source objects which have an attribute of `format` which exactly matches `urn:x-nmos:format:video` and a `device_id` attribute which exactly matches `9126cc2f-4c26-4c9b-a6cd-93c4381c9be5`.
 
 **Example 3: Querying Within Objects**
 
--- a/docs/2.5. APIs - Query Parameters.md
+++ b/docs/2.5. APIs - Query Parameters.md
@@ -8,34 +8,34 @@ The following document describes the expected usage and behaviour of these query
 
 ## Pagination
 
-Query APIs SHOULD support pagination of their API resources where the 'paged' trait is specified in the RAML documentation. A 501 HTTP status code MUST be returned where pagination is attempted against a Query API which does not implement it. Pagination is not used by WebSocket subscriptions.
+Query APIs SHOULD support pagination of their API resources where the `paged` trait is specified in the RAML documentation. A 501 HTTP status code MUST be returned where pagination is attempted against a Query API which does not implement it. Pagination is not used by WebSocket subscriptions.
 
-Query API clients MUST detect whether pagination is being used by examining the HTTP response headers for 'X-Paging-Limit' which MUST be returned in all cases where pagination is in use.
+Query API clients MUST detect whether pagination is being used by examining the HTTP response headers for `X-Paging-Limit` which MUST be returned in all cases where pagination is in use.
 
-The registry which backs the Query API SHOULD maintain a 'creation' and 'update' timestamp alongside each registered resource. These values SHOULD NOT be returned to API clients in the response body, but will be made available via headers and used as pagination cursors.
+The registry which backs the Query API SHOULD maintain a `creation` and `update` timestamp alongside each registered resource. These values SHOULD NOT be returned to API clients in the response body, but will be made available via headers and used as pagination cursors.
 
 To ensure that pagination does not result in resources being skipped, there SHOULD NOT be duplicate creation or update timestamps stored against resources of the same type (Node, Source, Flow, etc.). It is RECOMMENDED that these timestamps are stored with nanosecond resolution using a TAI timebase, which will allow clients to navigate collections based on a common understanding of time.
 
-The choice to page based on 'creation' or 'update' timestamp SHOULD depend on the client's intended use of the data:
+The choice to page based on `creation` or `update` timestamp SHOULD depend on the client's intended use of the data:
 
-- Paging based on 'creation' time from the start of a collection provides the best mechanism to identify all resources held in a registry.
-- Paging based on 'update' time from the end of a collection provides the best mechanism to watch for changing resources in the case that WebSocket subscriptions are not available.
+- Paging based on `creation` time from the start of a collection provides the best mechanism to identify all resources held in a registry.
+- Paging based on `update` time from the end of a collection provides the best mechanism to watch for changing resources in the case that WebSocket subscriptions are not available.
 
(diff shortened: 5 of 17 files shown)


docs/2.1. APIs - Common Keys.md
```diff
@@ -72,33 +72,33 @@ A URN describing the role which a Device has within the production environment (
...
 ## Use of URNs
 
-All Uniform Resource Names (URNs) specified within the scope of the NMOS specifications MUST use the namespace 'urn:x-nmos:'. Manufacturers MAY use their own namespaces to indicate values of URN which are not currently defined within the NMOS namespace.
+All Uniform Resource Names (URNs) specified within the scope of the NMOS specifications MUST use the namespace `urn:x-nmos:`. Manufacturers MAY use their own namespaces to indicate values of URN which are not currently defined within the NMOS namespace.
 
-Where a URN begins 'urn:x-nmos:' it MUST use the following construction:
+Where a URN begins `urn:x-nmos:` it MUST use the following construction:
 
-- The URN 'base' is a sequence of one or more names delimited by ':' characters, including the namespace
+- The URN `base` is a sequence of one or more names delimited by `:` characters, including the namespace
```
garethsb:
```suggestion
- The URN 'base' is a sequence of one or more names delimited by `:` characters, including the namespace
```
Or just leave the word base unquoted...

docs/2.5. APIs - Query Parameters.md
```diff
@@ -8,34 +8,34 @@ The following document describes the expected usage and behaviour of these query
 
 ## Pagination
 
-Query APIs SHOULD support pagination of their API resources where the 'paged' trait is specified in the RAML documentation. A 501 HTTP status code MUST be returned where pagination is attempted against a Query API which does not implement it. Pagination is not used by WebSocket subscriptions.
+Query APIs SHOULD support pagination of their API resources where the `paged` trait is specified in the RAML documentation. A 501 HTTP status code MUST be returned where pagination is attempted against a Query API which does not implement it. Pagination is not used by WebSocket subscriptions.
```
garethsb: Not sure about the trait name paged being back-ticked?

docs/2.5. APIs - Query Parameters.md
```diff
@@ -8,34 +8,34 @@ The following document describes the expected usage and behaviour of these query
...
+Paging through a collection for a second time, using the same cursors as previously SHOULD NOT result in new data appearing in the response payloads when using `creation` ordering, provided the requested paging times are less than or equal to the maximum paging time held in the registry. However, note that as the registry is dynamic and resources could be updated or deleted a given cursor MAY return fewer results than it did previously. When using `update` ordering and a query filter, new data MAY appear when cursors are used for a second time.
 
 When query parameters which perform filtering are used at the same time as paging, the filters MUST be applied by the implementation before applying paging parameters to the resulting data set.
 
 A server MAY choose its own default value for paging limit (see Example 1).
 
-Where both 'since' and 'until' parameters are specified, the 'since' value takes precedence where a resulting data set is constrained by the server's value of 'limit' (see Example 5).
+Where both `since` and `until` parameters are specified, the `since` value takes precedence where a resulting data set is constrained by the server`s value of `limit` (see Example 5).
```
garethsb:
```suggestion
Where both `since` and `until` parameters are specified, the `since` value takes precedence where a resulting data set is constrained by the server's value of `limit` (see Example 5).
```

docs/2.5. APIs - Query Parameters.md
```diff
@@ -8,34 +8,34 @@ The following document describes the expected usage and behaviour of these query
...
-Servers SHOULD include 'prev' and 'next' Link headers and MAY include 'first' and 'last' Link headers in all responses.
+Servers SHOULD include `prev` and `next` Link headers and MAY include `first` and `last` Link headers in all responses.
 
 ### Examples
 
-The following examples show pagination for a set of registered data. In order to avoid displaying full resource representations, the only data listed here is the 'update' timestamp associated with each registered record. The same procedures can be applied where 'creation' timestamps and the '?paging.order=create' parameter are used instead.
+The following examples show pagination for a set of registered data. In order to avoid displaying full resource representations, the only data listed here is the `update` timestamp associated with each registered record. The same procedures can be applied where `creation` timestamps and the `?paging.order=create` parameter are used instead.
 
-Where a paging limit is not specified in a request the server's default is used.
+Where a paging limit is not specified in a request the server`s default is used.
```
garethsb:
```suggestion
Where a paging limit is not specified in a request the server's default is used.
```

docs/2.5. APIs - Query Parameters.md
```diff
@@ -88,8 +88,13 @@ Payload Resources
 ***Notes***
 
 - The data set returned when no query parameters are specified MUST be from the most recently updated (or created) resources in the collection, returned in descending order.
-- The X-Paging headers identify properties of the collection of data returned in the response. These parameters can be used to construct a URL which would return the same set of bounded data on consecutive requests (for example /x-nmos/query/v1.1/nodes?paging.since=0:10&paging.until=0:20).
-- The Link header identifies the 'next' and 'prev' cursors which an application MAY use to make its next requests. An implementation MAY also provide 'first' and 'last' cursors to identify the beginning and end of a collection if this can be addressed via a consistent URL.
+- The `X-Paging` headers identify properties of the collection of data returned in the response. These parameters can be used to construct a URL which would return the same set of bounded data on consecutive requests, for example:
```
garethsb: Just a suggestion...

```suggestion
- The `X-Paging-` headers identify properties of the collection of data returned in the response. These parameters can be used to construct a URL which would return the same set of bounded data on consecutive requests, for example:
```

docs/2.5. APIs - Query Parameters.md
```diff
@@ -238,13 +243,13 @@ Payload Resources
 
 ***Notes***
 
-- Whilst both 'since' and 'until' are specified, as this server example has a default paging limit of '10', the 'since' parameter  takes precedence. As a result of this the value of X-Paging-Until is lower than requested and a further request MUST be made to retrieve any remaining data.
+- Whilst both `since` and `until` are specified, as this server example has a default paging limit of `10`, the `since` parameter  takes precedence. As a result of this the value of X-Paging-Until is lower than requested and a further request MUST be made to retrieve any remaining data.
```
garethsb:
```suggestion
- Whilst both `since` and `until` are specified, as this server example has a default paging limit of `10`, the `since` parameter  takes precedence. As a result of this the value of `X-Paging-Until` is lower than requested and a further request MUST be made to retrieve any remaining data.
```

docs/3.0. Discovery.md
```diff
@@ -24,7 +24,7 @@ When performing a DNS-SD browse, clients SHOULD proceed as follows:
 
 1. Identify whether the client has been configured with DNS server addresses and a default search domain either manually or automatically via DHCP.
 2. If such configuration exists and the discovery mechanism is not explicitly set to mDNS, perform a unicast DNS browse for services of the appropriate type via the search domain.
-3. If no unicast responses are received and the discovery mechanism is not explicitly set to unicast DNS, perform a multicast DNS browse for services of the appropriate type in the '.local' domain.
+3. If no unicast responses are received and the discovery mechanism is not explicitly set to unicast DNS, perform a multicast DNS browse for services of the appropriate type in the  `.local ` domain.
```
garethsb:
```suggestion
3. If no unicast responses are received and the discovery mechanism is not explicitly set to unicast DNS, perform a multicast DNS browse for services of the appropriate type in the `.local` domain.
```

docs/3.1. Discovery - Registered Operation.md
```diff
@@ -8,42 +8,42 @@ This document describes usage of NMOS APIs for discovery in cases where where a
 
 ### DNS-SD Advertisement
 
-The preferred method of Registration API advertisement is via unicast DNS-SD advertisement of the type \_nmos-register.\_tcp. Registration APIs MUST additionally be capable of producing an mDNS advertisement. This MAY be disabled via a user-configurable method.
+The preferred method of Registration API advertisement is via unicast DNS-SD advertisement of the type  `_nmos-register._tcp`. Registration APIs MUST additionally be capable of producing an mDNS advertisement. This MAY be disabled via a user-configurable method.
```
garethsb:
```suggestion
The preferred method of Registration API advertisement is via unicast DNS-SD advertisement of the type `_nmos-register._tcp`. Registration APIs MUST additionally be capable of producing an mDNS advertisement. This MAY be disabled via a user-configurable method.
```

docs/3.1. Discovery - Registered Operation.md
```diff
@@ -8,42 +8,42 @@ This document describes usage of NMOS APIs for discovery in cases where where a
...
 
-Registration APIs supporting versions 'v1.2' and below MUST additionally be capable of producing an mDNS advertisement of type \_nmos-registration.\_tcp. This MAY be disabled via a user-configurable method.
+Registration APIs supporting versions `v1.2` and below MUST additionally be capable of producing an mDNS advertisement of type `_nmos-registration._tcp`. This MAY be disabled via a user-configurable method.
 
-*Note: RFC6763 Section 7.2 specifies that the maximum service name length for an mDNS advertisement is 16 characters when including the leading underscore, but "_nmos-registration" is 18 characters.*
+*Note: RFC6763 Section 7.2 specifies that the maximum service name length for an mDNS advertisement is 16 characters when including the leading underscore, but "`_nmos-registration`" is 18 characters.*
```
garethsb: Just a couple of suggestions...
```suggestion
*Note: [RFC 6763 Section 7.2](https://tools.ietf.org/html/rfc6763#section-7.2) specifies that the maximum service name length for an mDNS advertisement is 16 characters when including the leading underscore, but `_nmos-registration` is 18 characters.*
```

docs/1.0. Overview.md
```diff
@@ -70,9 +70,9 @@ Registered discovery takes place using a **Registration & Discovery System (RDS)
 
 ![Registration and Discovery](images/registration-and-discovery.png)
 
-The Registration Service implements the Registration API of the NMOS Discovery and Registration Specification. Nodes POST to this API to register themselves and their resources. The Registration Service also manages garbage collection of Nodes and their resources by requiring Nodes to send regular keep-alive/heartbeat messages.
+The Registration Service implements the Registration API of the NMOS Discovery and Registration Specification. Nodes `POST` to this API to register themselves and their resources. The Registration Service also manages garbage collection of Nodes and their resources by requiring Nodes to send regular keep-alive/heartbeat messages.
```
garethsb, line 73: We probably ought to write down rules that explain why we backtick certain terms. (I think it's probably reasonable here as something that appears in a machine-readable message format.

docs/2.0. APIs.md
```diff
@@ -29,20 +29,21 @@ All APIs MUST provide a JSON representation signalled via 'Content-Type: applica
...
 
+```
 http(s)://<ip address or hostname>:<port>/x-nmos/<api type>/<api version>/
 ```
 
 At each level of the API from the base resource upwards, the response SHOULD include a JSON format array of the child resources available from that point.
 
-API clients using DNS-SD for API discovery SHOULD construct these paths by making use of the TXT records 'api\_proto' and 'api\_ver', along with addresses and ports resolved via DNS-SD. API clients which are discovering Node APIs via a Query API SHOULD construct Node API paths using the corresponding data available within the \'api\' attributes within the Query API /nodes/ path.
+API clients using DNS-SD for API discovery SHOULD construct these paths by making use of the TXT records 'api\_proto' and 'api\_ver', along with addresses and ports resolved via DNS-SD. API clients which are discovering Node APIs via a Query API SHOULD construct Node API paths using the corresponding data available within the `api` attributes within the Query API `/nodes/` path.
```
garethsb: Did you decide not these two for some reason? ...Also, what about TXT?

```suggestion
API clients using DNS-SD for API discovery SHOULD construct these paths by making use of the TXT records `api_proto` and `api_ver`, along with addresses and ports resolved via DNS-SD. API clients which are discovering Node APIs via a Query API SHOULD construct Node API paths using the corresponding data available within the `api` attributes within the Query API `/nodes/` path.
```

docs/2.0. APIs.md
```diff
@@ -64,8 +65,8 @@ A v1.1 API response can include:
 ]
 ```
 
-- Appending /v1.0/ to the API base resource will request version 1.0 of the API if available.
-- The versioning format is v&lt;#MAJOR&gt;.&lt;#MINOR&gt;
+- Appending `/v1.0/` to the API base resource will request version 1.0 of the API if available.
+- The versioning format is `v<#MAJOR>.<#MINOR>`
 - MINOR increments will be performed for non-breaking changes (such as the addition of attributes in a response)
 - MAJOR increments will be performed for breaking changes (such as the renaming of a resource or attribute)
 - Versions MUST be represented as complete strings. Parsing MUST proceed as follows: separate into two strings, using the point (.) as a delimiter. Compare integer representations of MAJOR, MINOR version (such that v1.12 is greater than v1.5).
```
garethsb: Does this meet the (unwritten) rule or not?

Also why are MAJOR and MINOR capitalized here and not backticked? And why are the words MAJOR and MINOR preceded by a `#` above?

```suggestion
- Versions MUST be represented as complete strings. Parsing MUST proceed as follows: separate into two strings, using the point (`.`) as a delimiter. Compare integer representations of MAJOR, MINOR version (such that v1.12 is greater than v1.5).
```

docs/3.1. Discovery - Registered Operation.md
```diff
@@ -8,42 +8,42 @@ This document describes usage of NMOS APIs for discovery in cases where where a
...
 
-The DNS-SD advertisement MUST be accompanied by a TXT record of name 'api\_auth' with a value of either 'true' or 'false' dependent on whether authorization is needed for Registration API requests.
+The DNS-SD advertisement MUST be accompanied by a TXT record of name `api_auth` with a value of either `true` or `false` dependent on whether authorization is needed for Registration API requests.
 
-### pri
+#### `pri`
 
-The DNS-SD advertisement MUST include a TXT record with key 'pri' and an integer value. Servers MAY additionally represent a matching priority via the DNS-SD SRV record 'priority' and 'weight' as defined in RFC 2782. The TXT record SHOULD be used in favour to the SRV priority and weight where these values differ in order to overcome issues in the Bonjour and Avahi implementations.
+The DNS-SD advertisement MUST include a TXT record with key `pri` and an integer value. Servers MAY additionally represent a matching priority via the DNS-SD SRV record `priority` and `weight` as defined in RFC 2782. The TXT record SHOULD be used in favour to the SRV priority and weight where these values differ in order to overcome issues in the Bonjour and Avahi implementations.
```
garethsb: Link to RFC 2782 or not? Probably no need as we're saying it's optional?

docs/3.1. Discovery - Registered Operation.md
```diff
@@ -61,36 +61,36 @@ If no Registration APIs are advertised on a network, the Node SHOULD assume peer
...
+The preferred method of Query API advertisement is via unicast DNS-SD advertisement of the type `_nmos-query._tcp`. Query APIs MUST additionally be capable of producing an mDNS advertisement. This MAY be disabled via a user-configurable method.
 
 The IP address and port of the Query API MUST be identified via the DNS-SD advertisement, with the full HTTP path then being resolved via the standard NMOS API path documentation.
 
 Multiple DNS-SD advertisements for the same API are permitted where the API is exposed via multiple ports and/or protocols.
 
 ### DNS-SD TXT Records
 
-**api\_proto**
+**api_proto**
```
garethsb:
```suggestion
#### `api_proto`
```

docs/3.1. Discovery - Registered Operation.md
```diff
@@ -61,36 +61,36 @@ If no Registration APIs are advertised on a network, the Node SHOULD assume peer
...
 
-The DNS-SD advertisement MUST be accompanied by a TXT record of name 'api\_proto' with a value of either 'http' or 'https' dependent on the protocol in use by the Query API web server.
+The DNS-SD advertisement MUST be accompanied by a TXT record of name `api_proto` with a value of either `http` or `https` dependent on the protocol in use by the Query API web server.
 
-**api\_ver**
+**api_ver**
```
garethsb:
```suggestion
#### `api_ver`
```

docs/3.1. Discovery - Registered Operation.md
```diff
@@ -61,36 +61,36 @@ If no Registration APIs are advertised on a network, the Node SHOULD assume peer
...
 
-The DNS-SD advertisement MUST be accompanied by a TXT record of name 'api\_ver'. The value of this TXT record is a comma separated list of API versions supported by the server. For example: 'v1.0,v1.1,v2.0'. There SHOULD be no whitespace between commas, and versions SHOULD be listed in ascending order.
+The DNS-SD advertisement MUST be accompanied by a TXT record of name `api_ver`. The value of this TXT record is a comma separated list of API versions supported by the server. For example: `v1.0,v1.1,v2.0`. There SHOULD be no whitespace between commas, and versions SHOULD be listed in ascending order.
 
-**api\_auth**
+**api_auth**
```
garethsb:
```suggestion
#### `api_auth`
```

docs/3.1. Discovery - Registered Operation.md
```diff
@@ -61,36 +61,36 @@ If no Registration APIs are advertised on a network, the Node SHOULD assume peer
...
 
-The DNS-SD advertisement MUST be accompanied by a TXT record of name 'api\_auth' with a value of either 'true' or 'false' dependent on whether authorization is needed in order to interact with the Query API or not.
+The DNS-SD advertisement MUST be accompanied by a TXT record of name `api_auth` with a value of either `true` or `false` dependent on whether authorization is needed in order to interact with the Query API or not.
 
 **pri**
```
garethsb:
```suggestion
#### `pri`
```

docs/3.1. Discovery - Registered Operation.md
```diff
@@ -61,36 +61,36 @@ If no Registration APIs are advertised on a network, the Node SHOULD assume peer
...
 
-The DNS-SD advertisement MUST include a TXT record with key 'pri' and an integer value. Servers MAY additionally represent a matching priority via the DNS-SD SRV record 'priority' and 'weight' as defined in RFC 2782. The TXT record SHOULD be used in favour to the SRV priority and weight where these values differ in order to overcome issues in the Bonjour and Avahi implementations.
+The DNS-SD advertisement MUST include a TXT record with key `pri` and an integer value. Servers MAY additionally represent a matching priority via the DNS-SD SRV record `priority` and `weight` as defined in RFC 2782. The TXT record SHOULD be used in favour to the SRV priority and weight where these values differ in order to overcome issues in the Bonjour and Avahi implementations.
```
garethsb: RFC 2782 again.

docs/3.1. Discovery - Registered Operation.md
```diff
@@ -61,36 +61,36 @@ If no Registration APIs are advertised on a network, the Node SHOULD assume peer
...
 Values 0 to 99 correspond to an active NMOS Query API (zero being the highest priority). Values 100+ are reserved for development work to avoid colliding with a live system.
 
 ### Client Interaction Procedure
 
 1. Node (or control interface) comes online
 
-2. Node scans for an active Query API on the network using unicast and/or multicast DNS service discovery (type '\_nmos-query.\_tcp') as described in the [Discovery](3.0.%20Discovery.md#unicast-vs-multicast-dns-sd) document.
+2. Node scans for an active Query API on the network using unicast and/or multicast DNS service discovery (type `_nmos-query._tcp`) as described in the [Discovery](3.0.%20Discovery.md#unicast-vs-multicast-dns-sd) document.
 
 3. Given multiple returned Query APIs, the Node orders these based on their advertised priority (TXT pri), filtering out any APIs which do not support the desired API version, protocol and authorization mode (TXT api_ver, api_proto and api_auth).
```
garethsb: Yes or no?

```suggestion
3. Given multiple returned Query APIs, the Node orders these based on their advertised priority (TXT `pri`), filtering out any APIs which do not support the desired API version, protocol and authorization mode (TXT `api_ver`, `api_proto` and `api_auth`).
```

docs/3.1. Discovery - Registered Operation.md
```diff
@@ -8,42 +8,42 @@ This document describes usage of NMOS APIs for discovery in cases where where a
...
 
 1. Node comes online
 
-2. Node scans for an active Registration API on the network using unicast and/or multicast DNS service discovery (type '\_nmos-register.\_tcp') as described in the [Discovery](3.0.%20Discovery.md#unicast-vs-multicast-dns-sd) document.
+2. Node scans for an active Registration API on the network using unicast and/or multicast DNS service discovery (type `_nmos-register._tcp`) as described in the [Discovery](3.0.%20Discovery.md#unicast-vs-multicast-dns-sd) document.
 
-   - Nodes which support v1.2 and earlier versions MUST additionally browse for the deprecated service type '\_nmos-registration.\_tcp'.
+   - Nodes which support v1.2 and earlier versions MUST additionally browse for the deprecated service type `_nmos-registration._tcp`.
 
 3. Given multiple returned Registration APIs, the Node orders these based on their advertised priority (TXT pri), filtering out any APIs which do not support the desired API version, protocol and authorization mode (TXT api_ver, api_proto and api_auth).
```
garethsb: Yes or no?

```suggestion
3. Given multiple returned Registration APIs, the Node orders these based on their advertised priority (TXT `pri`), filtering out any APIs which do not support the desired API version, protocol and authorization mode (TXT `api_ver`, `api_proto` and `api_auth`).
```

docs/3.2. Discovery - Peer to Peer Operation.md
```diff
@@ -12,11 +12,11 @@ When configured for peer-to-peer discovery, a Node MUST advertise its presence w
 
 In the absence of a network registry, a Node MUST also indicate the validity of its API resources with mDNS TXT records.
 
-Both of the above are properties of the [Node API](../APIs/NodeAPI.raml) and require no additional implementation steps.
+`Both of the above are properties of the [Node API](../APIs/NodeAPI.raml) and require no additional implementation steps.
```
garethsb:
```suggestion
Both of the above are properties of the [Node API](../APIs/NodeAPI.raml) and require no additional implementation steps.
```

docs/3.2. Discovery - Peer to Peer Operation.md
```diff
@@ -26,43 +26,43 @@ From v1.3, Node mDNS announcements SHOULD only be used for peer to peer mode as
...
-| ver\_flw            | flows                               |
-| ver\_dvc            | devices                             |
-| ver\_snd            | senders                             |
-| ver\_rcv            | receivers                           |
+| `ver_slf`            | `self`                                |
+| `ver_src`            | `sources`                             |
+| `ver_flw`            | `flows`                               |
+| `ver_dvc`            | `devices`                             |
+| `ver_snd`            | `senders`                             |
+| `ver_rcv`            | `receivers`                           |
```
garethsb: Hmm, backtick the "Corresponding Node API Resource" or not? I guess if it's viewed as a component of a machine-read/writable URL path then yes. If it's the human name, perhaps it should be capitalized, Sources, etc. instead?

Elsewhere you extended the other rows of tables to match the new length?

docs/3.2. Discovery - Peer to Peer Operation.md
```diff
@@ -26,43 +26,43 @@ From v1.3, Node mDNS announcements SHOULD only be used for peer to peer mode as
...
 
-The value of each of the above SHOULD be an unsigned 8-bit integer initialised to '0'. This integer MUST be incremented and mDNS TXT record updated whenever a change is made to the corresponding HTTP API resource on the Node. The integer MUST wrap back to a value of '0' after reaching a maximum value of '255' (MAX_UINT8_T).
+The value of each of the above SHOULD be an unsigned 8-bit integer initialised to `0`. This integer MUST be incremented and mDNS TXT record updated whenever a change is made to the corresponding HTTP API resource on the Node. The integer MUST wrap back to a value of `0` after reaching a maximum value of `255` (MAX_UINT8_T).
```
garethsb, line 54: If these are the text value of the TXT (or `TXT`...) records, then backticking them makes sense. If they're just integers, does it? I might argue for just removing the quotes here...

docs/3.2. Discovery - Peer to Peer Operation.md
```diff
@@ -26,43 +26,43 @@ From v1.3, Node mDNS announcements SHOULD only be used for peer to peer mode as
...
 
-For example, the 'ver_src' TXT record MUST be created when the Node first advertises itself via mDNS. If the data held within the HTTP resource for /sources is added to, removed from or edited, then the 'ver_src' text record MUST be modified (value incremented).
+For example, the `ver_src` TXT record MUST be created when the Node first advertises itself via mDNS. If the data held within the HTTP resource for `/sources` is added to, removed from or edited, then the `ver_src` text record MUST be modified (value incremented).
```
garethsb, line 56: Ah, interesting, here `sources` _is_ prefixed by a slash to indicate its an HTTP URL path component.

docs/4.1. Behaviour - Registration.md
```diff
@@ -16,7 +16,7 @@ The following behaviour assumes a Node with a single network interface for all t
 2. The Node runs an HTTP accessible Node API.
 3. The Node produces an mDNS advertisement of type '\_nmos-node.\_tcp' in the '.local' domain as specified in [Node API](../APIs/NodeAPI.raml).
```
garethsb:
```suggestion
3. The Node produces an mDNS advertisement of type `_nmos-node._tcp` in the `.local` domain as specified in [Node API](../APIs/NodeAPI.raml).
```

docs/4.1. Behaviour - Registration.md
```diff
@@ -16,7 +16,7 @@ The following behaviour assumes a Node with a single network interface for all t
...
 4. The Node performs a DNS-SD browse for services of type '\_nmos-register.\_tcp' as specified in [Discovery: Registered Operation](3.1.%20Discovery%20-%20Registered%20Operation.md).
```
garethsb:
```suggestion
4. The Node performs a DNS-SD browse for services of type `_nmos-register._tcp` as specified in [Discovery: Registered Operation](3.1.%20Discovery%20-%20Registered%20Operation.md).
```

docs/4.1. Behaviour - Registration.md
```diff
@@ -16,7 +16,7 @@ The following behaviour assumes a Node with a single network interface for all t
...
-5. The Node registers itself with the Registration API by taking the object it holds under the Node API's /self resource and POSTing this to the Registration API.
+5. The Node registers itself with the Registration API by taking the object it holds under the Node API's /self resource and `POST`ing this to the Registration API.
```
garethsb:
```suggestion
5. The Node registers itself with the Registration API by taking the object it holds under the Node API's `/self` resource and `POST`ing this to the Registration API.
```

docs/4.1. Behaviour - Registration.md
```diff
@@ -16,7 +16,7 @@ The following behaviour assumes a Node with a single network interface for all t
...
 6. The Node persists itself in the registry by issuing heartbeats as below.
 7. The Node registers its other resources (from /devices, /sources etc) with the Registration API. Resources MUST be registered in the correct order, such that a Receiver which references a device_id is registered after that corresponding Device (for example).
```
garethsb:
```suggestion
7. The Node registers its other resources (from `/devices`, `/sources`, etc.) with the Registration API. Resources MUST be registered in the correct order, such that a Receiver which references a `device_id` is registered after that corresponding Device (for example).
```

Where do we stop...

docs/4.1. Behaviour - Registration.md
```diff
@@ -57,11 +57,11 @@ A Node SHOULD register resources in the following order to avoid rejections due
 
 (\*) Version v1.0 Flows do not have a device_id and SHOULD be garbage collected based on their parent source_id.
```
garethsb: I can't make a suggestion for the lines above this which mention `node_id`, etc.

```suggestion
(\*) Version v1.0 Flows do not have a `device_id` and SHOULD be garbage collected based on their parent `source_id`.
```

Also, somewhere above, it says "version 1.0", whereas here it says "Version v1.0". Which is right?

docs/4.1. Behaviour - Registration.md
```diff
@@ -81,13 +81,13 @@ The following error conditions describe likely scenarios encountered by Nodes wh
...
+If a Node is restarted, its first action upon discovering a Registration API is to register its 'Node' resource. On first registration with a Registration API this SHOULD result in a '201 Created' HTTP response code. If a Node receives a 200 code in this case, a previous record of the Node can be assumed to still exist within the network registry. In order to avoid the registry-held representation of the Node's resources from being out of sync with the Node's view, an HTTP `DELETE` SHOULD be performed in this situation to explicitly clear the registry of the Node and any sub-resources. A new Node registration after this point SHOULD result in the correct 201 response code.
```
garethsb: Why is the word Node in single quotes in the first sentence?
Should it be `201 Created` with backticks in the second sentence?
What about "a `200` code"??

docs/4.1. Behaviour - Registration.md
```diff
@@ -81,13 +81,13 @@ The following error conditions describe likely scenarios encountered by Nodes wh
...
 
 ### Node Encounters HTTP 400 (Or Other Undocumented 4xx) On Registration
 
 A 400 error indicates a client error which is likely to be the result of a validation failure identified by the Registration API. The same request MUST NOT be re-attempted without corrective action being taken first.
 
-A registry MAY issue a 400 code from the `/resource` POST endpoint for various reasons, including the following cases. It is not expected that a Node will be able to automatically handle these errors and user intervention could be necessary.
+A registry MAY issue a 400 code from the `/resource` `POST` endpoint for various reasons, including the following cases. It is not expected that a Node will be able to automatically handle these errors and user intervention could be necessary.
```
garethsb: Again here and below, is "400" an integer or a message fragment?

docs/4.1. Behaviour - Registration.md
```diff
@@ -111,6 +111,6 @@ A 500 error, inability to connect or a timeout indicates a server side or connec
...
 
-Note that in order to achieve efficient failover from one Registration API to another, the Node needs to perform its first heartbeat with the new Registration API within the garbage collection interval following its most recent successful heartbeat. It is therefore advised that Nodes use the heartbeat interval to define their connection timeouts for `/health/nodes/{nodeId}` POST requests.
+Note that in order to achieve efficient failover from one Registration API to another, the Node needs to perform its first heartbeat with the new Registration API within the garbage collection interval following its most recent successful heartbeat. It is therefore advised that Nodes use the heartbeat interval to define their connection timeouts for `/health/nodes/{nodeId}` `POST` requests.
 
 If a 5xx error is encountered when interacting with all discoverable Registration APIs, clients SHOULD implement an exponential backoff algorithm in their next attempts until a non-5xx response code is received.
```
garethsb, line 116: (And "5xx" should follow the same policy.)

docs/4.2. Behaviour - Querying.md
```diff
@@ -10,25 +10,25 @@ Where the behaviour associated with an API attribute is not sufficiently clear f
 
 ### Subscriptions
 
-`secure` Indicates whether the WebSocket connection is provided via an encrypted connection (ws:// vs. wss://). Unless otherwise indicated the value of this attribute SHOULD be 'false' if the API is being presented via HTTP, and 'true' for HTTPS. Query API clients MAY choose to specify the opposite value in requests for subscriptions, however they will receive a 400 response code unless the Query API explicitly supports a mismatch between encrypted HTTP and WebSocket connections.
+`secure` Indicates whether the WebSocket connection is provided via an encrypted connection (`ws://` vs. `wss://`). Unless otherwise indicated the value of this attribute SHOULD be `false` if the API is being presented via HTTP, and `true` for HTTPS. Query API clients MAY choose to specify the opposite value in requests for subscriptions, however they will receive a 400 response code unless the Query API explicitly supports a mismatch between encrypted HTTP and WebSocket connections.
```
garethsb: Either a colon after "`secure`", or a lowercase "i" on "Indicates"?. And in the next paragraph for `authorization` too...

**Edit:** looks like that other similar paragraphs below have the colon and keep the capital.

docs/4.2. Behaviour - Querying.md
```diff
@@ -10,25 +10,25 @@ Where the behaviour associated with an API attribute is not sufficiently clear f
...
 
 WebSockets created for specific clients will be cleaned up automatically if they disconnect.
 
 Connection upgrade procedures permitting an existing HTTP query to transition to a WebSocket are not currently supported, but could be in the future.
 
 ### Query API Behaviour
 
-Long-lived connections to the Query API are supported via WebSockets which can be set up via the /subscriptions resource. Query APIs MAY additionally support the HTTP 'Upgrade' header sent by clients to upgrade an HTTP GET request to a WebSocket. In cases where this is performed, a corresponding entry in /subscriptions MUST also be created with matching query parameters.
+Long-lived connections to the Query API are supported via WebSockets which can be set up via the `/subscriptions` resource. Query APIs MAY additionally support the HTTP `Upgrade` header sent by clients to upgrade an HTTP GET request to a WebSocket. In cases where this is performed, a corresponding entry in `/subscriptions` MUST also be created with matching query parameters.
```
garethsb: If we're backticking HTTP verbs...

```suggestion
Long-lived connections to the Query API are supported via WebSockets which can be set up via the `/subscriptions` resource. Query APIs MAY additionally support the HTTP `Upgrade` header sent by clients to upgrade an HTTP `GET` request to a WebSocket. In cases where this is performed, a corresponding entry in `/subscriptions` MUST also be created with matching query parameters.
```

docs/4.3. Behaviour - Nodes.md
```diff
@@ -8,44 +8,44 @@ Where the behaviour associated with an API attribute is not sufficiently clear f
...
 
-`interface_bindings`: For push-based transports the Sender 'interface_bindings' SHOULD contain a single network interface unless a redundancy mechanism such as ST.2022-7 is in use, in which case each 'leg' SHOULD have its matching interface listed. Where the redundancy mechanism sends more than one copy of the stream via the same interface, that interface SHOULD be listed a corresponding number of times. For pull-based transports the Sender 'interface_bindings' SHOULD list each interface which the Sender is listening for connections on.
+`interface_bindings`: For push-based transports the Sender `interface_bindings` SHOULD contain a single network interface unless a redundancy mechanism such as ST.2022-7 is in use, in which case each 'leg' SHOULD have its matching interface listed. Where the redundancy mechanism sends more than one copy of the stream via the same interface, that interface SHOULD be listed a corresponding number of times. For pull-based transports the Sender `interface_bindings` SHOULD list each interface which the Sender is listening for connections on.
```
garethsb:
```suggestion
`interface_bindings`: For push-based transports the Sender `interface_bindings` SHOULD contain a single network interface unless a redundancy mechanism such as ST 2022-7 is in use, in which case each 'leg' SHOULD have its matching interface listed. Where the redundancy mechanism sends more than one copy of the stream via the same interface, that interface SHOULD be listed a corresponding number of times. For pull-based transports the Sender `interface_bindings` SHOULD list each interface on which the Sender is listening for connections.
```

docs/4.3. Behaviour - Nodes.md
```diff
@@ -8,44 +8,44 @@ Where the behaviour associated with an API attribute is not sufficiently clear f
...
 
-`manifest_href`: The Sender 'manifest_href' SHOULD provide an HTTP(S) accessible URL to a file describing how to connect to the Sender (SDP in the case of RTP). The Sender's 'version' attribute SHOULD be updated if the contents of this file are modified. This URL MAY return an HTTP 404 where the 'active' parameter in the 'subscription' object is present and set to false (v1.2+ only). The value SHOULD be `null` when the transport type used by the Sender does not require a transport file.
+`manifest_href`: The Sender `manifest_href` SHOULD provide an HTTP(S) accessible URL to a file describing how to connect to the Sender (SDP in the case of RTP). The Sender's `version` attribute SHOULD be updated if the contents of this file are modified. This URL MAY return an HTTP 404 where the `active` parameter in the `subscription` object is present and set to false (v1.2+ only). The value SHOULD be `null` when the transport type used by the Sender does not require a transport file.
```
garethsb: Are we backticking Boolean values, right?

```suggestion
`manifest_href`: The Sender `manifest_href` SHOULD provide an HTTP(S) accessible URL to a file describing how to connect to the Sender (SDP in the case of RTP). The Sender's `version` attribute SHOULD be updated if the contents of this file are modified. This URL MAY return an HTTP 404 where the `active` parameter in the `subscription` object is present and set to `false` (v1.2+ only). The value SHOULD be `null` when the transport type used by the Sender does not require a transport file.
```

docs/4.3. Behaviour - Nodes.md
```diff
@@ -8,44 +8,44 @@ Where the behaviour associated with an API attribute is not sufficiently clear f
...
 
-`interface_bindings`: For push-based transports the Receiver 'interface_bindings' SHOULD contain a single network interface unless a redundancy mechanism such as ST.2022-7 is in use, in which case each 'leg' SHOULD have its matching interface listed. Where the redundancy mechanism receives more than one copy of the stream via the same interface, that interface SHOULD be listed a corresponding number of times. For push-based transports the Receiver 'interface_bindings' SHOULD list a single interface through which content is being consumed.
+`interface_bindings`: For push-based transports the Receiver `interface_bindings` SHOULD contain a single network interface unless a redundancy mechanism such as ST.2022-7 is in use, in which case each 'leg' SHOULD have its matching interface listed. Where the redundancy mechanism receives more than one copy of the stream via the same interface, that interface SHOULD be listed a corresponding number of times. For push-based transports the Receiver `interface_bindings` SHOULD list a single interface through which content is being consumed.
```
garethsb:
```suggestion
`interface_bindings`: For push-based transports the Receiver `interface_bindings` SHOULD contain a single network interface unless a redundancy mechanism such as ST 2022-7 is in use, in which case each 'leg' SHOULD have its matching interface listed. Where the redundancy mechanism receives more than one copy of the stream via the same interface, that interface SHOULD be listed a corresponding number of times. For push-based transports the Receiver `interface_bindings` SHOULD list a single interface through which content is being consumed.
```

docs/5.1. Data Model - Identifier Mapping.md
```diff
@@ -22,26 +22,26 @@ All resources shown above are part of the JT-NM reference architecture. Each log
...
 A new Source is created in the following circumstances:
 
--   At the input of essence into a networked media enviroment from the 'edge' (ie. a legacy or incompatible protocol)
--   At the point of capture (ie. in a network connected camera or microphone)
--   At the mix point of two or more Flows of essence (ie. an audio mix bus, vision mix bus)
--   Where a multiplexed Source is split into multiple mono-essence Sources
--   Where a set of Flows are combined as a single multiplexed Source
+- At the input of essence into a networked media enviroment from the 'edge' (ie. a legacy or incompatible protocol)
+- At the point of capture (ie. in a network connected camera or microphone)
+- At the mix point of two or more Flows of essence (ie. an audio mix bus, vision mix bus)
```
garethsb:
```suggestion
- At the point of capture (i.e. in a network connected camera or microphone)
- At the mix point of two or more Flows of essence (i.e. an audio mix bus, vision mix bus)
```

docs/5.1. Data Model - Identifier Mapping.md
```diff
@@ -22,26 +22,26 @@ All resources shown above are part of the JT-NM reference architecture. Each log
...
+- Where a multiplexed Source is split into multiple mono-essence Sources
+- Where a set of Flows are combined as a single multiplexed Source
 
-Whenever a new Source is generated, its immediate parent Source(s) MUST be recorded. This is achieved via a 'parents' attribute in the Node API Sources resource. This contains an array of Source IDs corresponding to the Sources whose Flows have been brought together at this point. For example, a vision mixer which is fading between a Flow from Source A, and a Flow from Source B would indicate \["A", "B"\] in its 'parents' attribute for the duration of this transition.
+Whenever a new Source is generated, its immediate parent Source(s) MUST be recorded. This is achieved via a `parents` attribute in the Node API Sources resource. This contains an array of Source IDs corresponding to the Sources whose Flows have been brought together at this point. For example, a vision mixer which is fading between a Flow from Source A, and a Flow from Source B would indicate ["A", "B"\] in its `parents` attribute for the duration of this transition.
```
garethsb: I see a removed backslash before the open square bracket, but not before the close square bracket, is that right? But not backticking this? Hmm, equiv array below is, so suggestion:

```suggestion
Whenever a new Source is generated, its immediate parent Source(s) MUST be recorded. This is achieved via a `parents` attribute in the Node API Sources resource. This contains an array of Source IDs corresponding to the Sources whose Flows have been brought together at this point. For example, a vision mixer which is fading between a Flow from Source A, and a Flow from Source B would indicate `["<id of A>", "<id of B>"]` in its `parents` attribute for the duration of this transition.
```

docs/5.1. Data Model - Identifier Mapping.md
```diff
@@ -22,26 +22,26 @@ All resources shown above are part of the JT-NM reference architecture. Each log
...
 
 ### Flows
 
 A new Flow is created in the following circumstances:
 
--   Where a Flow passes through a processing element which is modelled as a new Source
--   Where a Flow is modified in a non-editorial manner (ie. audio volume changed, EQ applied, video contrast changed, video encoded or decoded)
+- Where a Flow passes through a processing element which is modelled as a new Source
+- Where a Flow is modified in a non-editorial manner (ie. audio volume changed, EQ applied, video contrast changed, video encoded or decoded)
```
garethsb:
```suggestion
- Where a Flow is modified in a non-editorial manner (i.e. audio volume changed, EQ applied, video contrast changed, video encoded or decoded)
```

docs/5.1. Data Model - Identifier Mapping.md
```diff
@@ -22,26 +22,26 @@ All resources shown above are part of the JT-NM reference architecture. Each log
...
 
 Sources and Flows MUST be advertised via a Node's API, and registered via the Registration API if available. This SHOULD only be done on the Device which originates the Source or Flow; if a Flow is received by a Device from the network it does not need to be re-advertised via APIs.
 
-Whenever a new Flow is generated, its immediate parent Flow(s) MUST be recorded. This is achieved via a 'parents' attribute in the Node API Flows resource. This contains an array of Flow IDs corresponding to the Flows which have been brought together at this point. For example, an encoder which codes a raw video Flow C into a coded video Flow D would indicate \["C"\] in its 'parents' attribute.
+Whenever a new Flow is generated, its immediate parent Flow(s) MUST be recorded. This is achieved via a `parents` attribute in the Node API Flows resource. This contains an array of Flow IDs corresponding to the Flows which have been brought together at this point. For example, an encoder which codes a raw video Flow C into a coded video Flow D would indicate `["C"]` in its `parents` attribute.
```
garethsb:
```suggestion
Whenever a new Flow is generated, its immediate parent Flow(s) MUST be recorded. This is achieved via a `parents` attribute in the Node API Flows resource. This contains an array of Flow IDs corresponding to the Flows which have been brought together at this point. For example, an encoder which codes a raw video Flow C into a coded video Flow D would indicate `["<id of C>"]` in its `parents` attribute.
```

docs/6.0. Upgrade Path.md
```diff
@@ -8,47 +8,47 @@ API versioning is specified in the [APIs](2.0.%20APIs.md) documentation, with pr
...
 
 Registrations MUST NOT proceed if the Node API version implemented does not exactly match the API version used by the Registration API.
 
-Where a Node implements version v1.2 or below, it MUST browse for both the \'\_nmos-register.\_tcp\' DNS-SD service type, and the legacy \'\_nmos-registration.\_tcp\' DNS-SD service type in order to retrieve the full list of available Registration APIs. De-duplication SHOULD be performed against this returned list.
+Where a Node implements version v1.2 or below, it MUST browse for both the `_nmos-register._tcp` DNS-SD service type, and the legacy `_nmos-registration._tcp` DNS-SD service type in order to retrieve the full list of available Registration APIs. De-duplication SHOULD be performed against this returned list.
```
garethsb, line 15: (Another instance of "version vX.Y".)

docs/6.0. Upgrade Path.md
```diff
@@ -8,47 +8,47 @@ API versioning is specified in the [APIs](2.0.%20APIs.md) documentation, with pr
...
 
 Nodes which support multiple versions simultaneously MUST ensure that all of their resources meet the schemas for each corresponding version of the specification which is supported. Where necessary features are only available in the most recent version(s) of the specification it could be necessary to expose only a limited subset of a Node's resources from lower versioned endpoints.
 
 ### v1.3 to v1.2
 
-*   Nodes: 'attached_network_device' (in 'interfaces'), 'authorization' (in 'api' 'endpoints'), 'authorization' (in 'services')
-*   Devices: 'authorization' (in 'controls')
-*   Sources: 'event_type'
-*   Flows: 'event_type'
+- Nodes: `attached_network_device` (in `interfaces`), `authorization` (in api endpoints), `authorization` (in `services`)
```
garethsb:
```suggestion
- Nodes: `attached_network_device` (in `interfaces`), `authorization` (in `api` `endpoints`), `authorization` (in `services`)
```

docs/6.0. Upgrade Path.md
```diff
@@ -60,29 +60,29 @@ It is strongly RECOMMENDED to match Query API client compatibility to the maximu
...
 ### Affected Keys From v1.3
 
-*   Devices: 'type' could be listed as any new variant of 'urn:x-nmos:device'
-*   Flows: 'transfer_characteristic' could be listed as any string to permit future extensibility
-*   Flows: 'colorspace' could be listed as any string to permit future extensibility
-*   Senders: 'transport' could be listed as any new variant of 'urn:x-nmos:transport'
-*   Senders: 'manifest_href' could be listed as `null` for transport types that do not require a transport file
-*   Receivers: 'transport' could be listed as any new variant of 'urn:x-nmos:transport'
+- Devices: `type` could be listed as any new variant of `urn:x-nmos:device`
+- Flows: `transfer_characteristic` could be listed as any string to permit future -ensibility
```
garethsb:
```suggestion
- Flows: `transfer_characteristic` could be listed as any string to permit future extensibility
```

docs/6.0. Upgrade Path.md
```diff
@@ -60,29 +60,29 @@ It is strongly RECOMMENDED to match Query API client compatibility to the maximu
...
+- Flows: `colorspace` could be listed as any string to permit future extensibility
+- Senders: `transport` could be listed as any new variant of `urn:x-nmos:transport`
+- Senders: `manifest_href` could be listed as `null` for transport types that do - require a transport file
```
garethsb:
```suggestion
- Senders: `manifest_href` could be listed as `null` for transport types that do not require a transport file
```

docs/6.0. Upgrade Path.md
```diff
@@ -60,29 +60,29 @@ It is strongly RECOMMENDED to match Query API client compatibility to the maximu
...
-* Upgrade Query API clients which support the ?query.downgrade parameter and as such can still access all data in the system. Check ahead of doing this that the Query API server implementation supports the downgrade feature.
-* Upgrade the registry and Registration API(s) to support registrations using the new version, whilst continuing to support registrations at the old version.
-* Upgrade Nodes in the system to register against the new API version. Any Nodes which also interact with the Query API will continue to query using the old version initially unless they support downgrade queries.
-* Once all relevant Nodes have been upgraded, all Query API clients can be upgraded or instructed to perform queries against the new version.
+- Upgrade the Query API(s) to support serving the new version of responses, whilst providing backwards compatibility to old versions active in the running system.
+- Upgrade Query API clients which support the ?query.downgrade parameter and as such can still access all data in the system. Check ahead of doing this that the Query API server implementation supports the downgrade feature.
```
garethsb: Backticks... but to keep the `?` or not? Hmm.

```suggestion
- Upgrade Query API clients which support the `query.downgrade` parameter and as such can still access all data in the system. Check ahead of doing this that the Query API server implementation supports the downgrade feature.
```

garethsb (changes requested):
First pass

docs/2.0. APIs.md
```diff
@@ -29,20 +29,21 @@ All APIs MUST provide a JSON representation signalled via 'Content-Type: applica
...
 ```
 
 At each level of the API from the base resource upwards, the response SHOULD include a JSON format array of the child resources available from that point.
 
-API clients using DNS-SD for API discovery SHOULD construct these paths by making use of the TXT records 'api\_proto' and 'api\_ver', along with addresses and ports resolved via DNS-SD. API clients which are discovering Node APIs via a Query API SHOULD construct Node API paths using the corresponding data available within the \'api\' attributes within the Query API /nodes/ path.
+API clients using DNS-SD for API discovery SHOULD construct these paths by making use of the TXT records 'api\_proto' and 'api\_ver', along with addresses and ports resolved via DNS-SD. API clients which are discovering Node APIs via a Query API SHOULD construct Node API paths using the corresponding data available within the `api` attributes within the Query API `/nodes/` path.
 
-Where protocol and versioning data is not available, clients MAY assume a v1.0 API, which operates via the 'http' protocol only.
+Where protocol and versioning data is not available, clients MAY assume a v1.0 API, which operates via the `http` protocol only.
```
garethsb: Or maybe just this?

```suggestion
Where protocol and versioning data is not available, clients MAY assume a v1.0 API, which operates via HTTP only.
```

docs/2.1. APIs - Common Keys.md
peterbrightwell: Unquoted best.

docs/2.5. APIs - Query Parameters.md
peterbrightwell: Let's discuss

docs/2.1. APIs - Common Keys.md
peterbrightwell:
```suggestion
- The URN base is a sequence of one or more names delimited by `:` characters, including the namespace
```

garethsb:
```suggestion
- The URN base is a sequence of one or more names delimited by `:` characters, including the namespace
```

docs/3.1. Discovery - Registered Operation.md
peterbrightwell: +1 to the first suggestion but the additional double quotes were deliberate as we are quoting something.

docs/3.1. Discovery - Registered Operation.md
peterbrightwell:
```suggestion
The DNS-SD advertisement MUST include a TXT record with key `pri` and an integer value. Servers MAY additionally represent a matching priority via the DNS-SD SRV record `priority` and `weight` (see RFC 2782). The TXT record SHOULD be used in favour to the SRV priority and weight where these values differ in order to overcome issues in the Bonjour and Avahi implementations.
```

docs/3.1. Discovery - Registered Operation.md
peterbrightwell:
```suggestion
The DNS-SD advertisement MUST include a TXT record with key `pri` and an integer value. Servers MAY additionally represent a matching priority via the DNS-SD SRV record `priority` and `weight` (see RFC 2782). The TXT record SHOULD be used in favour to the SRV priority and weight where these values differ in order to overcome issues in the Bonjour and Avahi implementations.
```

docs/3.2. Discovery - Peer to Peer Operation.md
peterbrightwell: I think we should backtick and will extend the row widths

docs/3.2. Discovery - Peer to Peer Operation.md
```diff
@@ -26,43 +26,43 @@ From v1.3, Node mDNS announcements SHOULD only be used for peer to peer mode as
...
-The DNS-SD advertisement MUST be accompanied by a TXT record of name 'api\_auth' with a value of either 'true' or 'false' dependent on whether authorization is needed for Node API requests.
+The DNS-SD advertisement MUST be accompanied by a TXT record of name `api_auth` with a value of either `true` or `false` dependent on whether authorization is needed for Node API requests.
 
-### ver\_
+### `ver_`
 
 When a Node is operating in peer-to-peer mode it MUST additionally advertise the following mDNS TXT records as part of its Node advertisement. If a Node is successfully registered with a Registration API it MUST withdraw advertisements of these TXT records. There is no requirement to register these TXT records with a unicast DNS service.
 
 | **TXT Record Name** | **Corresponding Node API Resource** |
 |---------------------|-------------------------------------|
```
peterbrightwell, line 46:
```suggestion
| **TXT Record Name** |  **Corresponding Node API Resource**  |
|---------------------|---------------------------------------|
```

docs/4.1. Behaviour - Registration.md
peterbrightwell: Node should be unquoted - I'll change.  Not sure about return codes... more recent specs haven't backticked them, but probably they should be.

peterbrightwell:
```suggestion
If a Node is restarted, its first action upon discovering a Registration API is to register its Node resource. On first registration with a Registration API this SHOULD result in a `201 Created` HTTP response code. If a Node receives a 200 code in this case, a previous record of the Node can be assumed to still exist within the network registry. In order to avoid the registry-held representation of the Node's resources from being out of sync with the Node's view, an HTTP `DELETE` SHOULD be performed in this situation to explicitly clear the registry of the Node and any sub-resources. A new Node registration after this point SHOULD result in the correct 201 response code.
```

docs/3.2. Discovery - Peer to Peer Operation.md
garethsb:
```suggestion
| `ver_slf`           | `/self`                             |
| `ver_src`           | `/sources`                          |
| `ver_flw`           | `/flows`                            |
| `ver_dvc`           | `/devices`                          |
| `ver_snd`           | `/senders`                          |
| `ver_rcv`           | `/receivers`                        |
```

docs/4.1. Behaviour - Registration.md
peterbrightwell: Let's discuss

docs/4.2. Behaviour - Querying.md
peterbrightwell: Colon + capital is good

docs/4.2. Behaviour - Querying.md
peterbrightwell: I decided to backtick verbs right at the end of yesterday's session, but don't feel strongly about them.

docs/4.2. Behaviour - Querying.md
garethsb:
```suggestion
`secure`: Indicates whether the WebSocket connection is provided via an encrypted connection (`ws://` vs. `wss://`). Unless otherwise indicated the value of this attribute SHOULD be `false` if the API is being presented via HTTP, and `true` for HTTPS. Query API clients MAY choose to specify the opposite value in requests for subscriptions, however they will receive a 400 response code unless the Query API explicitly supports a mismatch between encrypted HTTP and WebSocket connections.
```

docs/4.2. Behaviour - Querying.md
```diff
@@ -10,25 +10,25 @@ Where the behaviour associated with an API attribute is not sufficiently clear f
...
+`secure` Indicates whether the WebSocket connection is provided via an encrypted connection (`ws://` vs. `wss://`). Unless otherwise indicated the value of this attribute SHOULD be `false` if the API is being presented via HTTP, and `true` for HTTPS. Query API clients MAY choose to specify the opposite value in requests for subscriptions, however they will receive a 400 response code unless the Query API explicitly supports a mismatch between encrypted HTTP and WebSocket connections.
 
 `authorization` Indicates whether the WebSocket connection requires authorization in order to connect. Use of authorization is likely to be a deployment decision and be the same for all subscriptions exposed from a single Query API, and for the RESTful API itself. Query API clients MAY specify a requested value for this attribute, but in most circumstances this will result in a 400 response code if the Query API is operating in the opposite mode.
```
garethsb:
```suggestion
`authorization`: Indicates whether the WebSocket connection requires authorization in order to connect. Use of authorization is likely to be a deployment decision and be the same for all subscriptions exposed from a single Query API, and for the RESTful API itself. Query API clients MAY specify a requested value for this attribute, but in most circumstances this will result in a 400 response code if the Query API is operating in the opposite mode.
```

docs/4.2. Behaviour - Querying.md
```diff
@@ -10,25 +10,25 @@ Where the behaviour associated with an API attribute is not sufficiently clear f
...
 
 ## Creating a WebSocket subscription
 
 In order to connect to a WebSocket, a client MUST first request a subscription of a particular type and with particular query parameters by performing a POST as defined in the [Query API](../APIs/QueryAPI.raml) subscriptions documentation. Use of GET requests to find existing suitable WebSocket connections is strongly discouraged.
```
garethsb:
```suggestion
In order to connect to a WebSocket, a client MUST first request a subscription of a particular type and with particular query parameters by performing a `POST` as defined in the [Query API](../APIs/QueryAPI.raml) subscriptions documentation. Use of `GET` requests to find existing suitable WebSocket connections is strongly discouraged.
```

docs/4.2. Behaviour - Querying.md
```diff
@@ -10,25 +10,25 @@ Where the behaviour associated with an API attribute is not sufficiently clear f
...
 
-Upon receiving a request for a new subscription, the Query API SHOULD identify whether it already has any WebSockets open which provide for the requested query. If a WebSocket exists, the 'subscription' object representing it MAY be returned to the user. If a relevant WebSocket does not exist a new one SHOULD be created.
+Upon receiving a request for a new subscription, the Query API SHOULD identify whether it already has any WebSockets open which provide for the requested query. If a WebSocket exists, the `subscription` object representing it MAY be returned to the user. If a relevant WebSocket does not exist a new one SHOULD be created.
 
 In normal operation, once a WebSocket has no more clients subscribed to it the Query API MAY automatically remove it and its corresponding HTTP-advertised subscription object. The Query API MUST NOT acknowledge HTTP DELETE requests for WebSockets running in this 'non-persistent' mode, instead issuing an HTTP 403 response.
```
garethsb:
```suggestion
In normal operation, once a WebSocket has no more clients subscribed to it the Query API MAY automatically remove it and its corresponding HTTP-advertised subscription object. The Query API MUST NOT acknowledge HTTP `DELETE` requests for WebSockets running in this 'non-persistent' mode, instead issuing an HTTP 403 response.
```

docs/5.1. Data Model - Identifier Mapping.md
peterbrightwell: Strange, I remember changing that one already.

docs/6.0. Upgrade Path.md
peterbrightwell: oops

docs/6.0. Upgrade Path.md
peterbrightwell: OOPS!

docs/2.5. APIs - Query Parameters.md
```diff
@@ -238,13 +243,13 @@ Payload Resources
...
 
 ### Edge Cases
 
-When a client requests data which falls at the extreme ends of the stored data set it can be less clear what values SHOULD be returned in the 'X-Paging-Limit' and 'X-Paging-Since' headers. The following examples are intended to clarify these cases.
+When a client requests data which falls at the extreme ends of the stored data set it can be less clear what values SHOULD be returned in the `X-Paging-Limit` and `X-Paging-Since` headers. The following examples are intended to clarify these cases.
 
-Where a paging limit is not specified in a request the server's default is used.
+Where a paging limit is not specified in a request the server`s default is used.
```
garethsb:
```suggestion
Where a paging limit is not specified in a request the server's default is used.
```

docs/3.1. Discovery - Registered Operation.md
garethsb:
```suggestion
*Note: [RFC 6763 Section 7.2](https://tools.ietf.org/html/rfc6763#section-7.2) specifies that the maximum service name length for an mDNS advertisement is 16 characters when including the leading underscore, but "`_nmos-registration`" is 18 characters.*
```

docs/2.5. APIs - Query Parameters.md
garethsb: It's fine, I don't feel strongly.

docs/4.1. Behaviour - Registration.md
garethsb:
```suggestion
If a Node is restarted, its first action upon discovering a Registration API is to register its Node resource (`/self`). On first registration with a Registration API this SHOULD result in a 201 (Created) HTTP response code. If a Node receives a 200 code in this case, a previous record of the Node can be assumed to still exist within the network registry. In order to avoid the registry-held representation of the Node's resources from being out of sync with the Node's view, an HTTP `DELETE` SHOULD be performed in this situation to explicitly clear the registry of the Node and any sub-resources. A new Node registration after this point SHOULD result in the correct 201 response code.
```

docs/2.0. APIs.md
```diff
@@ -29,20 +29,21 @@ All APIs MUST provide a JSON representation signalled via 'Content-Type: applica
...
 
+```
 http(s)://<ip address or hostname>:<port>/x-nmos/<api type>/<api version>/
 ```
 
 At each level of the API from the base resource upwards, the response SHOULD include a JSON format array of the child resources available from that point.
 
-API clients using DNS-SD for API discovery SHOULD construct these paths by making use of the TXT records 'api\_proto' and 'api\_ver', along with addresses and ports resolved via DNS-SD. API clients which are discovering Node APIs via a Query API SHOULD construct Node API paths using the corresponding data available within the \'api\' attributes within the Query API /nodes/ path.
+API clients using DNS-SD for API discovery SHOULD construct these paths by making use of the TXT records `api_proto` and `api_ver`, along with addresses and ports resolved via DNS-SD. API clients which are discovering Node APIs via a Query API SHOULD construct Node API paths using the corresponding data available within the `api` attributes within the Query API `/nodes/` path.
```
```suggestion
API clients using DNS-SD for API discovery SHOULD construct these paths by making use of the TXT records `api_proto` and `api_ver`, along with addresses and ports resolved via DNS-SD. API clients which are discovering Node APIs via a Query API SHOULD construct Node API paths using the corresponding data available within the `api` attributes within the Query API `/nodes` resources.
```

docs/5.1. Data Model - Identifier Mapping.md
```diff
@@ -57,123 +57,122 @@ For physical Nodes, the Node ID MUST be universally unique to that Node, and rem
...
 
 It is suggested that Flow IDs be generated using the combination of:
 
--   The parent Source ID
--   A processing element type and index (or name) within the Device
--   A processing element Flow index (or name)\
--   Parent Flow ID\*\*
+- The parent Source ID
+- A processing element type and index (or name) within the Device
+- A processing element Flow index (or name)\
```
garethsb:
```suggestion
- A processing element Flow index (or name)
```

docs/5.1. Data Model - Identifier Mapping.md
```diff
@@ -212,18 +211,18 @@ Output
...
 Input
 
--   Fixed Receiver ID
+- Fixed Receiver ID
 
 Output
 
--   Flow ID generated based on input Flow ID (see *Source and Flow
+- Flow ID generated based on input Flow ID (see *Source and Flow
     Representation* above)
```
garethsb: Just consistent with below
```suggestion
- Flow ID generated based on input Flow ID (see *Source and Flow Representation* above)
```

docs/2.3. APIs - Server Side Implementation Notes.md
```diff
@@ -6,7 +6,7 @@ _(c) AMWA 2016, CC Attribution-NoDerivatives 4.0 International (CC BY-ND 4.0)_
 
 In order to permit web-based control interfaces to be hosted remotely, all NMOS APIs MUST implement valid CORS HTTP headers in responses to all requests.
 
-In addition to this, where highlighted in the API specifications servers MUST respond to HTTP pre-flight OPTIONS requests. Servers MAY additionally support HTTP OPTIONS requests made to any other API resource.
+In addition to this, where highlighted in the API specifications servers MUST respond to HTTP pre-flight `OPTIONS` requests. Servers MAY additionally support HTTP ``OPTIONS requests made to any other API resource.
```
garethsb:
```suggestion
In addition to this, where highlighted in the API specifications servers MUST respond to HTTP pre-flight `OPTIONS` requests. Servers MAY additionally support HTTP `OPTIONS` requests made to any other API resource.
```

docs/3.2. Discovery - Peer to Peer Operation.md
garethsb: Don't care strongly.

docs/4.1. Behaviour - Registration.md
garethsb: Let's _not_ add backticks around status codes.

docs/6.0. Upgrade Path.md
garethsb: Let's deal with the issues around saying "version vX.Y" or "version X.Y" and the hash preceding MAJOR and MINOR or not in the parent PR.

docs/1.0. Overview.md
garethsb: Let's do that once we've completed the parent PR.

PR #160, opened 2021-04-06
Backtick HTTP return codes

peterbrightwell:
E.g. `404` not 404.


docs/3.2. Discovery - Peer to Peer Operation.md
```diff
@@ -51,7 +51,7 @@ When a Node is operating in peer-to-peer mode it MUST additionally advertise the
 | `ver_snd`           | `/senders`                          |
 | `ver_rcv`           | `/receivers`                        |
 
-The value of each of the above SHOULD be an unsigned 8-bit integer initialised to `0`. This integer MUST be incremented and mDNS TXT record updated whenever a change is made to the corresponding HTTP API resource on the Node. The integer MUST wrap back to a value of `0` after reaching a maximum value of `255` (MAX_UINT8_T).
+The value of each of the above SHOULD be an unsigned 8-bit integer initialised to 0. This integer MUST be incremented and mDNS TXT record updated whenever a change is made to the corresponding HTTP API resource on the Node. The integer MUST wrap back to a value of 0 after reaching a maximum value of 255 (MAX_UINT8_T).
```
garethsb: Hmm, this now reads even less like it means the _text representation of_ an unsigned 8-bit integer? [RFC 6763 Section 6.5](https://tools.ietf.org/html/rfc6763#section-6.5) says the value is often ASCII or UTF-8 but can be opaque binary data. This probably needs a rewrite. :-/

docs/4.1. Behaviour - Registration.md
```diff
@@ -79,15 +79,15 @@ If a network cable is temporarily unplugged, dependent on the time for which it
...
+If a Node is restarted, its first action upon discovering a Registration API is to register its Node resource. On first registration with a Registration API this SHOULD result in a `201` (Created) HTTP response code. If a Node receives a `200` (OK) code in this case, a previous record of the Node can be assumed to still exist within the network registry. In order to avoid the registry-held representation of the Node's resources from being out of sync with the Node's view, an HTTP `DELETE` SHOULD be performed in this situation to explicitly clear the registry of the Node and any sub-resources. A new Node registration after this point SHOULD result in the correct `201` (Created) response code.
 
-### Node Encounters HTTP 400 (Or Other Undocumented 4xx) On Registration
+### Node Encounters HTTP `400` (Or Other Undocumented `4xx`) On Registration
```
garethsb, line 86: We should check whether the anchor tags in the rendered docs change as a result of this (e.g. getting an extra '-' in them).

docs/2.0. APIs.md
```diff
@@ -80,26 +80,26 @@ In order to overcome shortcomings in some common libraries, the following requir
 
 #### `GET` and `HEAD` Requests
 
-- Clients performing requests using these methods SHOULD correctly handle a 301 response (moved permanently).
-- When a 301 is supported, the client MUST follow the redirect in order to retrieve the response payload.
-- Servers implementing the APIs MUST support requests using these methods to both the trailing slash and non-trailing slash path to each resource. One of these MAY produce a 301 response causing a redirect to the other if preferred.
+- Clients performing requests using these methods SHOULD correctly handle a `301` response (moved permanently).
+- When a `301` is supported, the client MUST follow the redirect in order to retrieve the response payload.
+- Servers implementing the APIs MUST support requests using these methods to both the trailing slash and non-trailing slash path to each resource. One of these MAY produce a `301` response causing a redirect to the other if preferred.
```
```suggestion
- Clients performing requests using these methods SHOULD correctly handle a `301` (Moved Permanently) response.
- When a `301` (Moved Permanently) is supported, the client MUST follow the redirect in order to retrieve the response payload.
- Servers implementing the APIs MUST support requests using these methods to both the trailing slash and non-trailing slash path to each resource. One of these MAY produce a `301` (Moved Permanently) response causing a redirect to the other if preferred.
```

garethsb (changes requested):

docs/2.0. APIs.md
garethsb: Ah, you made the first change of these, and I wasn't sure about the repetition. (I missed these first time round because I only looked at 2, 4 and 5 codes. Oops.)

docs/3.2. Discovery - Peer to Peer Operation.md
garethsb: Suggestion... although, does something similar then need to be done to `pri` everywhere? Maybe it's not necessary... or maybe just deleting "unsigned 8-bit" instead of adding "the text representation of" would make it sound less like binary representation of the integer was meant...

```suggestion
The value of each of the above SHOULD be the text representation of an unsigned 8-bit integer initialised to 0. This integer MUST be incremented and mDNS TXT record updated whenever a change is made to the corresponding HTTP API resource on the Node. The integer MUST wrap back to a value of 0 after reaching a maximum value of 255.
```

garethsb (approved):
LGTM; the question it's raised in my mind about how `pri` is described, and the concern about existing links to the anchors whose titles now have backticks in them, can be addressed in the parent PR if necessary.

docs/4.1. Behaviour - Registration.md
peterbrightwell: Looks like it should be ok.  Local render gives

```
http://127.0.0.1:4000/is-04/branches/docs-language/docs/4.1._Behaviour_-_Registration.html#node-encounters-http-400-or-other-undocumented-4xx-on-registration
```

PR #162, opened 2021-04-14
Possible alternative language on Query API Behaviour

peterbrightwell:
This is a possible rewording based on a comment in the mega-conversation. Not sure if it's right though so take a look

### DIFF
--- a/docs/4.2. Behaviour - Querying.md
+++ b/docs/4.2. Behaviour - Querying.md
@@ -2,59 +2,51 @@
 
 _(c) AMWA 2016, CC Attribution-NoDerivatives 4.0 International (CC BY-ND 4.0)_
 
-The Query API provides a read-only interface to the Nodes and their sub-resources which are currently held in the registry. Browsing of the Query API can be performed using the RESTful API, or using a WebSocket subscription ([RFC 6455](https://tools.ietf.org/html/rfc6455)), which is suggested for most applications.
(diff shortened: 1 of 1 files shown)


garethsb:
I have a few minor quibbles with the rewording but there are bigger issues that could be resolved if we tackled the whole page. The previous/parent heading 'Creating a WebSocket subscription' already states some of what is under the 'Query API Behaviour' heading, and the whole is just misleading or at least confusing with respect to persistent vs non-persistent connections (a client is _required_ to specify the `persist` flag so there's no reason to talk primarily about one type and the other as an after-thought).

Are we OK to change the headings/fragment ids in a patch release? If so, I could try a slightly larger rewrite if you like?

peterbrightwell:
@garethsb That would be good. I think changing headings would still be editorial (unless proved otherwise) so ok for a point release.

garethsb:
One of the best references for the HTTP/1.1 Protocol Upgrade mechanism used for WebSocket is https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism, but perhaps that's more appropriate in a future INFO-xxx implementation guide than directly here in IS-04?

peterbrightwell:
@garethsb Your text in 27b43cc looks good.  @andrewbonney can you take a look also?

andrewbonney:
> @garethsb Your text in [27b43cc](https://github.com/AMWA-TV/nmos-discovery-registration/commit/27b43ccd9dfbd6d5b9da5473906ea9962423c0a2) looks good. @andrewbonney can you take a look also?

Looks good to me

PR #164, opened 2021-05-05
example.api.com -> api.example.com

garethsb:
As per https://tools.ietf.org/html/rfc2606#section-3, https://tools.ietf.org/html/rfc6761#section-6.5

IS-08, IS-09, IS-10, etc. already get this right. I suggest we make this change in IS-04 v1.3.1, IS-05 v1.1.1, and next editorial releases of other specs. (I'll also make a PR for AMWA-TV/nmos-template.)

### DIFF
--- a/APIs/NodeAPI.raml
+++ b/APIs/NodeAPI.raml
@@ -4,7 +4,7 @@
 # (c) AMWA 2016
 
 title: Node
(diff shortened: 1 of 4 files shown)


peterbrightwell (approved):
LGTM

PR #165, opened 2021-05-05
Add POST request examples for Subscriptions with query parameters

### LINKED ISSUE #122: Add more examples of using Query API subscription "params"
The doc for Query API query params has lots of examples, at https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.2.x/docs/2.5.%20APIs%20-%20Query%20Parameters.md#basic-queries.

The doc for Query WebSocket API explains parameters can be set up for the subscription using the `"params"`, but has no examples, at https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.2.x/docs/4.2.%20Behaviour%20-%20Querying.md#handling-query-parameters.

It would be nice to include some/all of the same examples to show how URL query parameters are mapped to `"params"`.

It would also be good to have the special `query.` params listed/linked from here as well - `query.rql`, `query.downgrade`, ...

garethsb:
Resolves #122. There's more that could be said (e.g. about `query.rql`, and to make clear that e.g. `paging.order` makes no sense) but at this stage, this may be enough.

### DIFF
--- a/docs/4.2. Behaviour - Querying.md
+++ b/docs/4.2. Behaviour - Querying.md
@@ -24,6 +24,8 @@ In order to connect to a WebSocket, a client first requests a Subscription of a 
 
 Upon receiving a request for a new Subscription, the Query API MAY return an existing Subscription to the user, if it matches the requested attributes. If a relevant Subscription does not exist, a new one SHOULD be created.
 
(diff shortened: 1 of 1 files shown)


peterbrightwell:
Nice

PR #168, opened 2021-05-27
Fix typo

### LINKED ISSUE #167: "push-based transports" and "pull-based transports"
Hi!

What is exactly the meaning of "push-based transports" and "pull-based transports" mentioned in the [node behaviour](https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.3.x/docs/4.3.%20Behaviour%20-%20Nodes.md)?

Also, in the next paragraph:

> For push-based transports the Receiver interface_bindings SHOULD contain a single network interface unless a redundancy mechanism such as ST 2022-7 is in use, in which case each 'leg' SHOULD have its matching interface listed. Where the redundancy mechanism receives more than one copy of the stream via the same interface, that interface SHOULD be listed a corresponding number of times. For push-based transports the Receiver interface_bindings SHOULD list a single interface through which content is being consumed.

"push-based transports" is mentioned in all cases, perhaps "pull-based transports" is missing?

Thanks in advance.

Victor.

garethsb:
Resolves #167, thanks, @VentuzVictor.


andrewbonney (approved):

peterbrightwell (approved):
You're right @garethsb this does make more sense!

PR #175, opened 2021-08-23
Remove document numbering; make list from README

peterbrightwell:
Markdown documents in `docs/` no longer use numbering for ordering and levelling, which is now specified in `docs/README.md` instead.  This does not change old versions, which still use numbering.  See also https://github.com/AMWA-TV/nmos-template/pull/25


andrewbonney (approved):
Personally I think it's a shame that the `docs` directory doesn't provide ordering anymore, but the changes appear fine.

PR #179, opened 2022-02-09
IS-04 Requirements for NMOS Controllers

jonathan-r-thorpe:
Addition of section for requirements specific to NMOS Controllers, based on the work done in the AWMA NMOS Controller Specification meetings.

### DIFF
--- a/docs/Controllers.md
+++ b/docs/Controllers.md
@@ -40,7 +40,7 @@ Versions MUST be represented as complete strings. Parsing MUST proceed as follow
 Implementers of Controllers are RECOMMENDED to support multiple versions of the NMOS APIs simultaneously in order to ease the upgrade process in live facilities.
 
 #### API Common Keys
-Controllers SHOULD follow the requirements for common APi keys specificified in the [IS-04 APIs: Common Keys](APIs%20-%20Common%20Keys.md) document including the requirements regarding [use of URNs](APIs%20-%20Common%20Keys.md#use-of-urns).
+Controllers SHOULD follow the requirements for common APi keys specified in the [IS-04 APIs: Common Keys](APIs%20-%20Common%20Keys.md) document including the requirements regarding [use of URNs](APIs%20-%20Common%20Keys.md#use-of-urns).
 
 ~~Later API versions might use URNs which have not yet been defined and so Controllers MUST be tolerant to these.~~
 
--- a/docs/Controllers.md
+++ b/docs/Controllers.md
@@ -62,18 +62,18 @@ In order to locate the Query API, the Controller SHOULD support all of the follo
 
 The Controller SHOULD offer unicast DNS-SD as the default mechanism. 
 
-Controllers SHOULD observe and interpret all of the TXT records returned with the DNS Service Discovery responses according to the requirements for Query API Clients specified in the [API Paths section of the APIs document](APIs.md#api-paths) in this specification.
+Controllers SHOULD observe and interpret all of the TXT records returned with the DNS Service Discovery responses according to the requirements for Query API Clients specified in the [APIs: API Paths](APIs.md#api-paths) section of this specification.
 
 ## Query API
 
 The Controller MUST be capable of using the Query API to discover any registered resource, including Node, Device, Source, Flow, Sender, and Receiver,
-as described in the [APIs section](README.md#apis) of this specification.
+as described in the [APIs](APIs.md) section of this specification.
 
 The Controller MUST use the Query API either via the HTTP API or by requesting WebSocket subscriptions.
 
 When using the Query API, query filters SHOULD be used (and advanced query language where available) to cut down on the volume of resources returned to the Controller, as specified in the [APIs: Query Parameters](APIs%20-%20Query%20Parameters.md) document.
 
-Controllers SHOULD adhere to the version downgrade requirements for Query API Clients specified in the [Requirements for Query API Clients section of the Upgrade Path document](Upgrade%20Path.md#requirements-for-query-api-clients) in this specification.
+Controllers SHOULD adhere to the version downgrade requirements for Query API Clients specified in the [Upgrade Path: Requirements for Query API Clients](Upgrade%20Path.md#requirements-for-query-api-clients) section of this specification.
 
 ## Pagination
 
@@ -82,7 +82,7 @@ This can be both hard to keep track of for a client, and require a large number 
 
 For this reason it is RECOMMENDED to use the RESTful QUERY API for debug and development purposes only.
 
-If using the HTTP API rather than WebSocket subscriptions, pagination requirements MUST be implemented as specified in the [Pagination section of the APIs: Query Parameters document](APIs%20-%20Query%20Parameters.md#pagination) in this specification.
+If using the HTTP API rather than WebSocket subscriptions, pagination requirements MUST be implemented as specified in the [APIs: Query Parameters: Pagination](APIs%20-%20Query%20Parameters.md#pagination) section of this specification.
 	
 ## WebSockets & Subscriptions	
 
--- a/docs/Controllers.md
+++ b/docs/Controllers.md
@@ -70,7 +70,7 @@ as described in the [APIs](APIs.md) section of this specification.
 
 When using the Query API, query filters SHOULD be used (and advanced query language where available) to cut down on the volume of resources returned to the Controller, as specified in the [APIs: Query Parameters](APIs%20-%20Query%20Parameters.md) document.
 
-Controllers SHOULD adhere to the version downgrade requirements for Query API Clients specified in the [Upgrade Path: Requirements for Query API Clients](Upgrade%20Path.md#requirements-for-query-api-clients) section of this specification.
+Controllers MUST adhere to the [Requirements for Query API Clients](Upgrade%20Path.md#requirements-for-query-api-clients) found in this specification.
 
 ## Pagination
 
@@ -82,7 +82,7 @@ In order to avoid polling of the HTTP API, it is RECOMMENDED that Controllers us
 
 ## Dynamic Update of Resources
 
-The Controller MUST be capable of using the Query API to discover and dynamically update the state of any registered resource, including Node, Device, Source, Flow, Sender, and Receiver.
+The Controller MUST be capable of using the Query API to discover and dynamically report the state of any registered resource, including Node, Device, Source, Flow, Sender, and Receiver.
 
 * The Controller MUST indicate available Senders to the user.
 * The Controller MUST reflect changes in presence/absence of Senders to the user after a maximum of 30 seconds.
--- a/docs/Controllers.md
+++ b/docs/Controllers.md
@@ -54,12 +54,15 @@ If a Controller receives an HTTP `5xx` or `4xx` response code from the API, a fa
 The Controller SHOULD display the content of the response's error field to the user if possible, and indicate that the resource is likely to be in an unknown state.
 The Controller SHOULD also refresh the endpoints of the relevant resources to ensure the Controller is accurately reflecting the current state of the API.
 
-## Registry Service Discovery	
+## Query API Discovery	
+
+NMOS [Discovery](Discovery.md) makes use of the DNS Service Discovery (DNS-SD) protocol. The [Query API: Client Interaction Procedure](Discovery%20-%20Registered%20Operation.md#client-interaction-procedure-1) section of this specification describes the requirements in detail.
 
 In order to locate the Query API, the Controller SHOULD support all of the following:
-* Unicast DNS Service Discovery (DNS-SD).
-* Multicast DNS Service Discovery (mDNS).
-* Direct configuration with the location of a preferred Query API.
+
+* Unicast DNS
+* Multicast DNS (mDNS)
+* Direct configuration with the location of a preferred Query API
 
 The Controller SHOULD offer unicast DNS-SD as the default mechanism. 
 
--- a/docs/Controllers.md
+++ b/docs/Controllers.md
@@ -71,7 +71,7 @@ This can be both hard to keep track of for a client, and require a large number 
 
 For this reason it is recommended to use the RESTful QUERY API for debug and development purposes only.
 
-If using the RESTful Query API rather than WebSockets, Pagination requirements MUST be implemented as specified in the [Pagination section of the APIs: Query Parameters document](APIs%20-%20Query%20Parameters.md#pagination) in this specification.
+If using the RESTful Query API rather than WebSockets, pagination requirements MUST be implemented as specified in the [Pagination section of the APIs: Query Parameters document](APIs%20-%20Query%20Parameters.md#pagination) in this specification.
 	
 ## WebSockets & Subscriptions	
 Where a WebSocket or other subscription based mechanism is provided for Controller usage, it is strongly recommended that Controllers make use of this and do not use the API resources directly.
--- a/docs/Controllers.md
+++ b/docs/Controllers.md
@@ -51,8 +51,8 @@ For Controllers performing `GET` and `HEAD` requests, using these methods SHOULD
 When a `301` is supported, the Controller MUST follow the redirect in order to retrieve the required response payload.
 
 If a Controller receives an HTTP `5xx` or `4xx` response code from the API, a failure has occurred.
-The Controller SHOULD display the content of the response's error field to the user if possible, and indicate that the resource is likely to be in an unknown state.
-The Controller SHOULD also refresh the endpoints of the relevant resources to ensure the Controller is accurately reflecting the current state of the API.
+The Controller SHOULD display the content of the response's `error` field to the user if possible.
+The Controller SHOULD continue indicating this error status for the resource until it has been refreshed.
 
 ## Query API Discovery	
 
--- a/docs/Controllers.md
+++ b/docs/Controllers.md
@@ -10,9 +10,11 @@ A Controller is Client software that interacts with the NMOS APIs to discover, c
 * This document covers how the Controller interacts with the NMOS APIs only.
 It does not cover other features of the Controller software, such as presentation.
 * This document does not cover any requirements relating to where a Controller is additionally acting as a Node (e.g. receiving monitoring information via IS-07).
-* Where this document refers to the "user" of a Controller, this includes both human operators who drive the Controller manually and automation systems that drive the Controller programmatically.
 
 ## General
+
+### User
+Where this document refers to the "user" of a Controller, this includes both human operators who drive the Controller manually and automation systems that drive the Controller programmatically.
 
 ### HTTP APIs
 #### Trailing Slashes
@@ -30,9 +32,9 @@ Later API versions might use URNs which have not yet been defined and so Control
 
 #### Error Codes & Responses
 The NMOS APIs use HTTP status codes to indicate success, failure and other cases to Controllers as per [RFC 7231](https://tools.ietf.org/html/rfc7231) and related standards.
-Where the RAML specification of an API specifies explicit response codes it is expected that a Controller will handle these cases in a particular way.
 
-As explicit handling of every possible HTTP response code is not expected, Controllers MUST instead implement more generic handling for ranges of response codes (1xx, 2xx, 3xx, 4xx and 5xx).
+As explicit handling of every possible HTTP response code is not expected, Controllers MUST implement generic handling for ranges of response codes (1xx, 2xx, 3xx, 4xx and 5xx).
+However, where the RAML specification of an API specifies explicit response codes the Controller SHOULD handle these cases explicitly.
 
 For Controllers performing GET and HEAD requests, using these methods SHOULD correctly handle a 301 (Moved Permanently) response.
 
--- a/docs/Controllers.md
+++ b/docs/Controllers.md
@@ -87,7 +87,7 @@ If using the HTTP API rather than WebSocket subscriptions, pagination requiremen
 	
 ## WebSockets & Subscriptions	
 
-Where a WebSocket or other subscription based mechanism is provided for Controller usage, it is strongly RECOMMENDED that Controllers make use of this and do not use the API resources directly.
+It is RECOMMENDED that Controllers make use of the WebSocket subscription mechanism and do not use the HTTP API resources directly.
 
 If a WebSocket connection fails, then an attempt to reconnect to the WebSocket SHOULD NOT be attempted. Instead, a new subscription SHOULD be created with this API or a different one if required. If all available APIs return errors, an exponential backoff algorithm SHOULD be used when retrying until a success code is returned.
 
--- a/docs/Controllers.md
+++ b/docs/Controllers.md
@@ -33,6 +33,8 @@ Controllers are responsible for identifying the correct API version they require
 Implementers of Controllers are strongly recommended to support multiple versions of the NMOS APIs simultaneously in order to ease the upgrade process in live facilities.
 ~~[reference](https://specs.amwa.tv/is-08/releases/v1.0.1/docs/5.0._Upgrade_Path.html#requirements-for-channel-mapping-clients)~~
 
+Controllers MUST be tolerant to URNs which have not yet been defined, but which might be added in later API versions as defined in the [Use of URNs section of the APIs: Common Keys document](APIs%20-%20Common%20Keys.md#use-of-urns) in this specification. (_JRT NOTE: moved into General section but is from IS-04 specification_)
+
 #### HTTP Error Codes & Responses
 The NMOS APIs use HTTP status codes to indicate success, failure and other cases to Controllers as per [RFC 7231](https://tools.ietf.org/html/rfc7231) and related standards.
 Where the RAML specification of an API specifies explicit response codes it is expected that a Controller will handle these cases in a particular way.
@@ -47,7 +49,7 @@ When a 301 is supported, the Controller MUST follow the redirect in order to ret
 ~~[reference](https://specs.amwa.tv/is-05/releases/v1.1.1/docs/2.0._APIs.html#urls-approach-to-trailing-slashes)~~
 
 If a Controller receives a HTTP 500 response code from the API, a failure has occurred. The Controller SHOULD display the content of the response’s error field to the user if possible, and indicate that the Device may be in a bad state.
-The Controller SHOULD also refresh the values in the map/active and map/activations endpoints to ensure it is accurately reflecting the current state of the API. (_JRT NOTE: This is very IS-08 specific, needs to be more generic_)
+The Controller SHOULD also refresh the values ~~in the map/active and map/activations endpoints~~ _of the relevant resources_ to ensure it is accurately reflecting the current state of the API. (_JRT NOTE: Was very IS-08 specific, modified to be more generic_)
 
 ~~[reference](https://specs.amwa.tv/is-08/releases/v1.0.1/docs/2.1._APIs_-_Controller_Side_Implementation.html#failure-modes)~~
 
@@ -60,7 +62,7 @@ In order to locate the Registry, the Controller SHOULD support all of the follow
 
 The Controller SHOULD offer unicast DNS-SD as the default mechanism. 
 
-Controllers SHOULD observe and interpret all of the TXT records returned with the DNS Service Discovery responses according to the requirements for Query API Clients specified in [the API Paths section of the APIs document in this specification](APIs.md#api-paths).
+Controllers SHOULD observe and interpret all of the TXT records returned with the DNS Service Discovery responses according to the requirements for Query API Clients specified in the [API Paths section of the APIs document](APIs.md#api-paths) in this specification.
 
 ~~These indicate the preferred API instance to use (via 'pri'), the API versions supported (via 'api_ver') and the protocol supported (via 'api_proto').~~
 
@@ -69,28 +71,25 @@ Controllers SHOULD observe and interpret all of the TXT records returned with th
 ~~[reference](https://specs.amwa.tv/is-04/releases/v1.3.1/docs/2.0._APIs.html#api-paths)~~
 ~~[reference](https://github.com/AMWA-TV/nmos/wiki/IS-04-Client)~~
 
-## API Version (_JRT NOTE: repetition of section name_)
-Controllers MUST be tolerant to URNs which have not yet been defined, but which might be added in later API versions. (_JRT NOTE: should this be part of the General section?_)
-
-Controllers SHOULD adhere to the version downgrade requirements for Query API Clients specified in the [Upgrade document in this specification](Upgrade%20Path.md#requirements-for-query-api-clients). (_JRT NOTE: move to the Query API section?_)
-
 ~~Make use of the Query API downgrade function wherever possible. This ensures that you can display Nodes and their resources via your interface, even if they only support a lower version of the API specifications.  This is described in the IS-04 specifcation~~
 
 ~~[reference](https://specs.amwa.tv/is-04/releases/v1.3.1/docs/6.0._Upgrade_Path.html#requirements-for-registries-registration-and-query-apis)
 [reference](https://github.com/AMWA-TV/nmos/wiki/IS-04-Client)~~
 
 ## Query API
-The Controller SHALL be capable of using the Registry's IS-04 Query API to discover any registered resource, including Node, Device,  Source, Flow, Sender, and Receiver.
+The Controller MUST be capable of using the Registry's IS-04 Query API to discover any registered resource, including Node, Device,  Source, Flow, Sender, and Receiver.
 
-~~The Controller SHALL use the Registry’s IS-04 Query API to discover all Senders or Receivers that are registered in the Registry.~~
+~~The Controller MUST use the Registry’s IS-04 Query API to discover all Senders or Receivers that are registered in the Registry.~~
 
-The Controller SHALL use the Registry’s IS-04 Query API either via the REST API or by requesting WebSocket subscriptions.
+The Controller MUST use the Registry’s IS-04 Query API either via the REST API or by requesting WebSocket subscriptions.
 
 When using the Query API, basic queries SHOULD be used (and advanced query language where available) to cut down on the volume of resources returned to the Controller.	
 ~~[reference](https://github.com/AMWA-TV/nmos/wiki/IS-04-Client)~~
 
-If using the RESTful API rather than WebSockets, Pagination requirements MUST be implemented as specified in the [APIs: Query Parameters document in this specification](APIs%20-%20Query%20Parameters.md#pagination).
+If using the RESTful API rather than WebSockets, Pagination requirements MUST be implemented as specified in the [Pagination section of the APIs: Query Parameters document](APIs%20-%20Query%20Parameters.md#pagination) in this specification.
 ~~[reference](https://specs.amwa.tv/is-04/releases/v1.3.1/docs/2.5._APIs_-_Query_Parameters.html#pagination)~~
+
+Controllers SHOULD adhere to the version downgrade requirements for Query API Clients specified in the [Requirements for Query API Clients section of the Upgrade Path document](Upgrade%20Path.md#requirements-for-query-api-clients) in this specification.
 
 ## WebSockets & Subscriptions	
 Where a WebSocket or other subscription based mechanism is provided for Controller usage, it is strongly recommended that Controllers make use of this and do not use the API resources directly.
@@ -107,10 +106,10 @@ If a WebSocket connection fails, then an attempt to reconnect to the WebSocket M
 ~~[reference](https://github.com/AMWA-TV/nmos/wiki/IS-04-Client)~~
 
 ## Dynamic Update of Resources
-The Controller SHALL be capable of using the Registry's IS-04 Query API to discover and dynamically update the state of any registered resource, including Node, Device,  Source, Flow, Sender, and Receiver.
+The Controller MUST be capable of using the Registry's IS-04 Query API to discover and dynamically update the state of any registered resource, including Node, Device,  Source, Flow, Sender, and Receiver.
 
-* The Controller SHALL indicate available Senders to the user.
-* The Controller SHALL dynamically indicate to the user when a Sender is put ‘offline’ or put back 'online'
-* ~~The Controller SHALL discover and dynamically update all the Senders that are registered in the Registry.~~
-* The Controller SHALL indicate to the user which of the discovered Receivers are controllable via the IS-05 Connection API, for instance, allowing Senders to be connected.
-* The Controller MUST indicate to the user which Receivers have a connection API, but can choose not to display Receivers without a connection API
+* The Controller MUST indicate available Senders to the user.
+* The Controller MUST dynamically indicate to the user when a Sender is put ‘offline’ or put back 'online'
+* ~~The Controller MUST discover and dynamically update all the Senders that are registered in the Registry.~~
+* The Controller MUST indicate to the user which of the discovered Receivers are controllable via the IS-05 Connection API, for instance, allowing Senders to be connected.
+* The Controller MUST indicate to the user which Receivers have a Connection API, but can choose not to display Receivers without a Connection API
--- a/docs/Controllers.md
+++ b/docs/Controllers.md
@@ -94,6 +94,6 @@ If a WebSocket connection fails, then an attempt to reconnect to the WebSocket S
 The Controller MUST be capable of using the Registry's IS-04 Query API to discover and dynamically update the state of any registered resource, including Node, Device,  Source, Flow, Sender, and Receiver.
 
 * The Controller MUST indicate available Senders to the user.
-* The Controller must reflect changes in presence/absence of Senders to the user after a maximum of 30 seconds.
+* The Controller MUST reflect changes in presence/absence of Senders to the user after a maximum of 30 seconds.
 * The Controller MUST indicate available Receivers to the user which have an IS-05 Connection API.
   However, the Controller can choose not to display discovered Receivers without an IS-05 Connection API.
--- a/docs/Controllers.md
+++ b/docs/Controllers.md
@@ -88,7 +88,7 @@ For this reason it is RECOMMENDED to use the RESTful QUERY API for debug and dev
 
 If using the HTTP API rather than WebSocket subscriptions, pagination requirements MUST be implemented as specified in the [APIs: Query Parameters: Pagination](APIs%20-%20Query%20Parameters.md#pagination) section of this specification.
 	
-## WebSockets & Subscriptions	
+## WebSocket Subscriptions	
 
 It is RECOMMENDED that Controllers make use of the WebSocket subscription mechanism and do not use the HTTP API resources directly.
 
--- a/docs/Controllers.md
+++ b/docs/Controllers.md
@@ -36,12 +36,12 @@ The NMOS APIs use HTTP status codes to indicate success, failure and other cases
 As explicit handling of every possible HTTP response code is not expected, Controllers MUST implement generic handling for ranges of response codes (1xx, 2xx, 3xx, 4xx and 5xx).
 However, where the RAML specification of an API specifies explicit response codes the Controller SHOULD handle these cases explicitly.
 
-For Controllers performing GET and HEAD requests, using these methods SHOULD correctly handle a 301 (Moved Permanently) response.
+For Controllers performing `GET` and `HEAD` requests, using these methods SHOULD correctly handle a `301` (Moved Permanently) response.
 
 When a 301 is supported, the Controller MUST follow the redirect in order to retrieve the required response payload.
 
 If a Controller receives a HTTP 500 response code from the API, a failure has occurred.
-The Controller SHOULD display the content of the response’s error field to the user if possible, and indicate that the Resource may be in an unknown state.
+The Controller SHOULD display the content of the responseâ€™s error field to the user if possible, and indicate that the Resource may be in an unknown state.
 The Controller SHOULD also refresh the endpoints of the relevant Resources to ensure the Controller is accurately reflecting the current state of the API.
 
 ## Registry Service Discovery	
@@ -59,7 +59,7 @@ Controllers SHOULD observe and interpret all of the TXT records returned with th
 The Controller MUST be capable of using the Registry's IS-04 Query API to discover any registered resource, including Node, Device,  Source, Flow, Sender, and Receiver,
 as described in the [APIs section](README.md#apis) of this specification.
 
-The Controller MUST use the Registry’s IS-04 Query API either via the REST API or by requesting WebSocket subscriptions.
+The Controller MUST use the Registryâ€™s IS-04 Query API either via the REST API or by requesting WebSocket subscriptions.
 
 When using the Query API, basic queries SHOULD be used (and advanced query language where available) to cut down on the volume of resources returned to the Controller.	
 
@@ -82,6 +82,6 @@ If a WebSocket connection fails, then an attempt to reconnect to the WebSocket S
 The Controller MUST be capable of using the Registry's IS-04 Query API to discover and dynamically update the state of any registered resource, including Node, Device,  Source, Flow, Sender, and Receiver.
 
 * The Controller MUST indicate available Senders to the user.
-* The Controller MUST dynamically indicate to the user when a Sender is put ‘offline’ or put back 'online'.
+* The Controller MUST dynamically indicate to the user when a Sender is put â€˜offlineâ€™ or put back 'online'.
 * The Controller MUST indicate available Receivers to the user which have an IS-05 Connection API.
 However, the Controller can choose not to display discovered Receivers without an IS-05 Connection API.
--- a/docs/Controllers.md
+++ b/docs/Controllers.md
@@ -42,15 +42,14 @@ Controllers SHOULD follow the requirements for common API keys specified in the 
 The NMOS APIs use HTTP status codes to indicate success, failure and other cases to Controllers as per [RFC 7231](https://tools.ietf.org/html/rfc7231) and related standards.
 
 As explicit handling of every possible HTTP response code is not expected, Controllers MUST implement generic handling for ranges of response codes (`1xx`, `2xx`, `3xx`, `4xx` and `5xx`).
-However, where the RAML specification of an API specifies explicit response codes the Controller SHOULD handle these cases explicitly.
+However, where the API specifies explicit response codes the Controller SHOULD handle these cases explicitly.
 
 For Controllers performing `GET` and `HEAD` requests, using these methods SHOULD correctly handle a `301` (Moved Permanently) response.
 
 When a `301` is supported, the Controller MUST follow the redirect in order to retrieve the required response payload.
 
 If a Controller receives an HTTP `5xx` or `4xx` response code from the API, a failure has occurred.
-The Controller SHOULD display the content of the response's `error` field to the user if possible.
-The Controller SHOULD continue indicating this error status for the resource until it has been refreshed.
+The Controller SHOULD display the content of the response's `error` field to the User if possible.
 
 ## Query API Discovery	
 
--- a/docs/Controllers.md
+++ b/docs/Controllers.md
@@ -19,13 +19,18 @@ It does not cover other features of the Controller software, such as presentatio
 
 ### HTTP APIs
 #### Trailing Slashes
-APIs may advertise URLs with or without a trailing slash. Controllers SHOULD exercise care when appending additional paths to these URLs.
+APIs may advertise URLs with or without a trailing slash.
+Controllers appending paths to `href` type attributes MUST support both cases, avoiding doubled or missing slashes.
+(~~[AMWA Wiki](https://github.com/AMWA-TV/nmos/wiki/Common-Issues)~~)
+(~~[IS-05](https://specs.amwa.tv/is-05/releases/v1.1.1/docs/2.0._APIs.html#client-behaviour)~~)
+
+~~APIs may advertise URLs with or without a trailing slash. Controllers SHOULD exercise care when appending additional paths to these URLs.~~
 (~~[AMWA Wiki](https://github.com/AMWA-TV/nmos/wiki/Common-Issues)~~)
 
 Controllers performing requests other than GET or HEAD (i.e PUT, POST, DELETE, OPTIONS etc.) MUST use URLs with no trailing slash present.
 (~~[IS-08](https://specs.amwa.tv/is-08/releases/v1.0.1/docs/2.0._APIs.html#all-other-requests-put-post-delete-options-etc)~~)
 
-When a server implementation needs to indicate an API URL via an href or similar attribute, it is valid to either include a trailing slash or not provided that the listed path is accessible and follows the above rules. Controllers appending paths to href type attributes MUST support both cases, avoiding doubled or missing slashes.
+~~When a server implementation needs to indicate an API URL via an href or similar attribute, it is valid to either include a trailing slash or not provided that the listed path is accessible and follows the above rules. Controllers appending paths to href type attributes MUST support both cases, avoiding doubled or missing slashes.~~
 (~~[IS-05](https://specs.amwa.tv/is-05/releases/v1.1.1/docs/2.0._APIs.html#client-behaviour)~~)
 
 #### API Version
--- a/docs/Controllers.md
+++ b/docs/Controllers.md
@@ -20,31 +20,26 @@ Where this document refers to the "user" of a Controller, this includes both hum
 
 #### Trailing Slashes
 
-~~APIs may advertise URLs with or without a trailing slash.
-Controllers appending paths to `href` type attributes MUST support both cases, avoiding doubled or missing slashes.~~
-
-
 Controllers appending paths to `href` type attributes MUST support URLs both with and without a trailing slash, to avoid doubled or missing slashes.
 
 Controllers performing requests other than `GET` or `HEAD` (i.e `PUT`, `POST`, `DELETE`, `OPTIONS` etc.) MUST use URLs with no trailing slash present.
 
 #### API Versions
+
 The versioning format is `v<MAJOR>.<MINOR>`
 * `MINOR` increments will be performed for non-breaking changes (such as the addition of attributes in a response)
 * `MAJOR` increments will be performed for breaking changes (such as the renaming of a resource or attribute)
 
 Versions MUST be represented as complete strings. Parsing MUST proceed as follows: separate into two strings, using the point (.) as a delimiter. Compare integer representations of `MAJOR`, `MINOR` version (such that v1.12 is greater than v1.5).
 
-~~Controllers are responsible for identifying the correct API version they require.~~
-
 Implementers of Controllers are RECOMMENDED to support multiple versions of the NMOS APIs simultaneously in order to ease the upgrade process in live facilities.
 
 #### API Common Keys
+
 Controllers SHOULD follow the requirements for common APi keys specified in the [IS-04 APIs: Common Keys](APIs%20-%20Common%20Keys.md) document including the requirements regarding [use of URNs](APIs%20-%20Common%20Keys.md#use-of-urns).
 
-~~Later API versions might use URNs which have not yet been defined and so Controllers MUST be tolerant to these.~~
-
 #### Error Codes & Responses
+
 The NMOS APIs use HTTP status codes to indicate success, failure and other cases to Controllers as per [RFC 7231](https://tools.ietf.org/html/rfc7231) and related standards.
 
 As explicit handling of every possible HTTP response code is not expected, Controllers MUST implement generic handling for ranges of response codes (1xx, 2xx, 3xx, 4xx and 5xx).
@@ -55,7 +50,7 @@ For Controllers performing `GET` and `HEAD` requests, using these methods SHOULD
 When a 301 is supported, the Controller MUST follow the redirect in order to retrieve the required response payload.
 
 If a Controller receives a HTTP 5xx or 4xx response code from the API, a failure has occurred.
-The Controller SHOULD display the content of the response's error field to the user if possible, and indicate that the resource is likely to ~~may~~ be in an unknown state.
+The Controller SHOULD display the content of the response's error field to the user if possible, and indicate that the resource is likely to be in an unknown state.
 The Controller SHOULD also refresh the endpoints of the relevant resources to ensure the Controller is accurately reflecting the current state of the API.
 
 ## Registry Service Discovery	
@@ -70,16 +65,18 @@ The Controller SHOULD offer unicast DNS-SD as the default mechanism.
 Controllers SHOULD observe and interpret all of the TXT records returned with the DNS Service Discovery responses according to the requirements for Query API Clients specified in the [API Paths section of the APIs document](APIs.md#api-paths) in this specification.
 
 ## Query API
+
 The Controller MUST be capable of using the Registry's IS-04 Query API to discover any registered resource, including Node, Device,  Source, Flow, Sender, and Receiver,
 as described in the [APIs section](README.md#apis) of this specification.
 
-The Controller MUST use the Registry's IS-04 Query API either via the ~~REST~~ HTTP API or by requesting WebSocket subscriptions.
+The Controller MUST use the Registry's IS-04 Query API either via the HTTP API or by requesting WebSocket subscriptions.
 
 When using the Query API, query filters SHOULD be used (and advanced query language where available) to cut down on the volume of resources returned to the Controller, as specified in the [APIs: Query Parameters](APIs%20-%20Query%20Parameters.md) document.
 
 Controllers SHOULD adhere to the version downgrade requirements for Query API Clients specified in the [Requirements for Query API Clients section of the Upgrade Path document](Upgrade%20Path.md#requirements-for-query-api-clients) in this specification.
 
 ## Pagination
+
 In large systems, API resources rely upon pagination in order to return high volumes of data.
 This can be both hard to keep track of for a client, and require a large number of requests in order to scan the entire data set.
 
@@ -88,6 +85,7 @@ For this reason it is RECOMMENDED to use the RESTful QUERY API for debug and dev
 If using the RESTful Query API rather than WebSockets, Pagination requirements MUST be implemented as specified in the [Pagination section of the APIs: Query Parameters document](APIs%20-%20Query%20Parameters.md#pagination) in this specification.
 	
 ## WebSockets & Subscriptions	
+
 Where a WebSocket or other subscription based mechanism is provided for Controller usage, it is strongly RECOMMENDED that Controllers make use of this and do not use the API resources directly.
 
 If a WebSocket connection fails, then an attempt to reconnect to the WebSocket SHOULD NOT be attempted. Instead, a new subscription SHOULD be created with this API or a different one if required. If all available APIs return errors, an exponential backoff algorithm SHOULD be used when retrying until a success code is returned.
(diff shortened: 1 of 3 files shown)


docs/Controllers.md
```diff
@@ -0,0 +1,85 @@
...
+#### API Version
+Controllers are responsible for identifying the correct API version they require.
+
+Implementers of Controllers are strongly recommended to support multiple versions of the NMOS APIs simultaneously in order to ease the upgrade process in live facilities.
+
+Later API versions might use URNs which have not yet been defined and so Controllers MUST be tolerant to these.
+
+#### Error Codes & Responses
+The NMOS APIs use HTTP status codes to indicate success, failure and other cases to Controllers as per [RFC 7231](https://tools.ietf.org/html/rfc7231) and related standards.
+Where the RAML specification of an API specifies explicit response codes it is expected that a Controller will handle these cases in a particular way.
```
prince-chrismc: Maybe expected can be replaced with normative language? 🤔 The next sentence seems to oppose this a little

jonathan-r-thorpe: Yes, good point. For reference, this section was lifted verbatim from [IS-08](https://specs.amwa.tv/is-08/releases/v1.0.1/docs/2.0._APIs.html#error-codes--responses) so same probably applies there.

jonathan-r-thorpe: I've updated the text in the latest commit

prince-chrismc (commented):
👍

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+## WebSockets & Subscriptions	
+Where a WebSocket or other subscription based mechanism is provided for Controller usage, it is strongly recommended that Controllers make use of this and do not use the API resources directly.
+
+If a WebSocket connection fails, then an attempt to reconnect to the WebSocket SHOULD NOT be attempted. Instead, a new subscription SHOULD be created with this API or a different one if required. If all available APIs return errors, an exponential backoff algorithm SHOULD be used when retrying until a success code is returned.
+
+## Dynamic Update of Resources
+The Controller MUST be capable of using the Registry's IS-04 Query API to discover and dynamically update the state of any registered resource, including Node, Device,  Source, Flow, Sender, and Receiver.
+
+* The Controller MUST indicate available Senders to the user.
+* The Controller MUST dynamically indicate to the user when a Sender is put �offline� or put back 'online'.
```
alabou: Do we have a definition for ‘offline’ and 'online ? Does it mean on the network ... active ... master enabled ...

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+Controllers appending paths to `href` type attributes MUST support both cases, avoiding doubled or missing slashes.
+
+Controllers performing requests other than GET or HEAD (i.e PUT, POST, DELETE, OPTIONS etc.) MUST use URLs with no trailing slash present.
+
+#### API Version
+Controllers are responsible for identifying the correct API version they require.
+
+Implementers of Controllers are strongly recommended to support multiple versions of the NMOS APIs simultaneously in order to ease the upgrade process in live facilities.
+
+Later API versions might use URNs which have not yet been defined and so Controllers MUST be tolerant to these.
```
garethsb: Not just URNs but other values also, as discussed in IS-04 already.

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+
+#### Error Codes & Responses
+The NMOS APIs use HTTP status codes to indicate success, failure and other cases to Controllers as per [RFC 7231](https://tools.ietf.org/html/rfc7231) and related standards.
+
+As explicit handling of every possible HTTP response code is not expected, Controllers MUST implement generic handling for ranges of response codes (1xx, 2xx, 3xx, 4xx and 5xx).
+However, where the RAML specification of an API specifies explicit response codes the Controller SHOULD handle these cases explicitly.
+
+For Controllers performing GET and HEAD requests, using these methods SHOULD correctly handle a 301 (Moved Permanently) response.
```
garethsb: I believe IS-04 now generally highlights HTTP verbs and response codes in `backticks`.

```suggestion
For Controllers performing `GET` and `HEAD` requests, using these methods SHOULD correctly handle a `301` (Moved Permanently) response.
```

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+
+When a 301 is supported, the Controller MUST follow the redirect in order to retrieve the required response payload.
+
+If a Controller receives a HTTP 500 response code from the API, a failure has occurred.
+The Controller SHOULD display the content of the response�s error field to the user if possible, and indicate that the Resource may be in an unknown state.
```
garethsb: Please avoid lowercase conformance keywords like `may` if at all possible, and change directional quotes and apostrophes in the Markdown source like `’` for straight ones like `'`.

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+The NMOS APIs use HTTP status codes to indicate success, failure and other cases to Controllers as per [RFC 7231](https://tools.ietf.org/html/rfc7231) and related standards.
+
+As explicit handling of every possible HTTP response code is not expected, Controllers MUST implement generic handling for ranges of response codes (1xx, 2xx, 3xx, 4xx and 5xx).
+However, where the RAML specification of an API specifies explicit response codes the Controller SHOULD handle these cases explicitly.
+
+For Controllers performing GET and HEAD requests, using these methods SHOULD correctly handle a 301 (Moved Permanently) response.
+
+When a 301 is supported, the Controller MUST follow the redirect in order to retrieve the required response payload.
+
+If a Controller receives a HTTP 500 response code from the API, a failure has occurred.
```
garethsb: Why only `500` itself rather than any `5xx` or indeed `4xx` codes?

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+The Controller SHOULD display the content of the response�s error field to the user if possible, and indicate that the Resource may be in an unknown state.
+The Controller SHOULD also refresh the endpoints of the relevant Resources to ensure the Controller is accurately reflecting the current state of the API.
+
+## Registry Service Discovery	
+
+In order to locate the Registry, the Controller SHOULD support all of the following: 
+* Unicast DNS Service Discovery (DNS-SD).
+* mDNS.
```
garethsb:
```suggestion
* Multicast DNS Service Discovery (mDNS).
```

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+* Direct configuration with the location of a preferred Registry.
+
+The Controller SHOULD offer unicast DNS-SD as the default mechanism. 
+
+Controllers SHOULD observe and interpret all of the TXT records returned with the DNS Service Discovery responses according to the requirements for Query API Clients specified in the [API Paths section of the APIs document](APIs.md#api-paths) in this specification.
+
+## Query API
+The Controller MUST be capable of using the Registry's IS-04 Query API to discover any registered resource, including Node, Device,  Source, Flow, Sender, and Receiver,
```
garethsb: I see `resource` uncapitalized here but capitalized as `Resource` above. Not sure whether the rest of IS-04 is consistent on this, probably not...

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+as described in the [APIs section](README.md#apis) of this specification.
+
+The Controller MUST use the Registry�s IS-04 Query API either via the REST API or by requesting WebSocket subscriptions.
+
+When using the Query API, basic queries SHOULD be used (and advanced query language where available) to cut down on the volume of resources returned to the Controller.
```
garethsb: I don't think `basic queries` is clear enough, as a simple `GET /nodes` request is a basic query, but I think you mean SHOULD take advantage of the filtering capabilities, right?

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+## Registry Service Discovery	
+
+In order to locate the Registry, the Controller SHOULD support all of the following: 
+* Unicast DNS Service Discovery (DNS-SD).
+* mDNS.
+* Direct configuration with the location of a preferred Registry.
+
+The Controller SHOULD offer unicast DNS-SD as the default mechanism. 
+
+Controllers SHOULD observe and interpret all of the TXT records returned with the DNS Service Discovery responses according to the requirements for Query API Clients specified in the [API Paths section of the APIs document](APIs.md#api-paths) in this specification.
```
garethsb: Here you have a link to the existing normative language, but above about error codes and responses you don't. Please can we (initially at least) have links everywhere so we can judge how much of this is _new_ requirements vs. aggregation/summarisation of existing requirements?

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+
+When using the Query API, basic queries SHOULD be used (and advanced query language where available) to cut down on the volume of resources returned to the Controller.	
+
+Controllers SHOULD adhere to the version downgrade requirements for Query API Clients specified in the [Requirements for Query API Clients section of the Upgrade Path document](Upgrade%20Path.md#requirements-for-query-api-clients) in this specification.
+
+## Pagination
+In large systems, API resources rely upon pagination in order to return high volumes of data.
+This can be both hard to keep track of for a client, and require a large number of requests in order to scan the entire data set.
+
+For this reason it is recommended to use the RESTful QUERY API for debug and development purposes only.
```
garethsb: If that's normative, you want capitalized `RECOMMENDED`.
However, I personally think this recommendation only applies "if the client needs the entire data set locally".

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+
+The Controller MUST use the Registry�s IS-04 Query API either via the REST API or by requesting WebSocket subscriptions.
+
+When using the Query API, basic queries SHOULD be used (and advanced query language where available) to cut down on the volume of resources returned to the Controller.	
+
+Controllers SHOULD adhere to the version downgrade requirements for Query API Clients specified in the [Requirements for Query API Clients section of the Upgrade Path document](Upgrade%20Path.md#requirements-for-query-api-clients) in this specification.
+
+## Pagination
+In large systems, API resources rely upon pagination in order to return high volumes of data.
+This can be both hard to keep track of for a client, and require a large number of requests in order to scan the entire data set.
```
garethsb: Dispute "hard to keep track of".

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+
+For this reason it is recommended to use the RESTful QUERY API for debug and development purposes only.
+
+If using the RESTful Query API rather than WebSockets, Pagination requirements MUST be implemented as specified in the [Pagination section of the APIs: Query Parameters document](APIs%20-%20Query%20Parameters.md#pagination) in this specification.
```
garethsb:
```suggestion
If using the HTTP API rather than WebSocket subscriptions, pagination requirements MUST be implemented as specified in the [Pagination section of the APIs: Query Parameters document](APIs%20-%20Query%20Parameters.md#pagination) in this specification.
```

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+	
+## WebSockets & Subscriptions	
+Where a WebSocket or other subscription based mechanism is provided for Controller usage, it is strongly recommended that Controllers make use of this and do not use the API resources directly.
```
garethsb: Please capitalize `recommended` to make clear this is conformance language.
And somewhere here or in accompanying documentation, explain the justification.

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+
+If a WebSocket connection fails, then an attempt to reconnect to the WebSocket SHOULD NOT be attempted. Instead, a new subscription SHOULD be created with this API or a different one if required. If all available APIs return errors, an exponential backoff algorithm SHOULD be used when retrying until a success code is returned.
```
garethsb: Why not attempt a reconnection to the same WebSocket??

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Dynamic Update of Resources
+The Controller MUST be capable of using the Registry's IS-04 Query API to discover and dynamically update the state of any registered resource, including Node, Device,  Source, Flow, Sender, and Receiver.
```
garethsb: I take "IS-04 Query API" to mean the HTTP API. And yet the above text strongly recommends against using it...

docs/Controllers.md
garethsb: ~Agree it's unclear. Also "dynamically".~ Resolved

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+#### Trailing Slashes
+APIs may advertise URLs with or without a trailing slash.
+Controllers appending paths to `href` type attributes MUST support both cases, avoiding doubled or missing slashes.
+
+Controllers performing requests other than GET or HEAD (i.e PUT, POST, DELETE, OPTIONS etc.) MUST use URLs with no trailing slash present.
+
+#### API Version
+Controllers are responsible for identifying the correct API version they require.
+
+Implementers of Controllers are strongly recommended to support multiple versions of the NMOS APIs simultaneously in order to ease the upgrade process in live facilities.
```
garethsb: If that's normative, you want capitalized `RECOMMENDED`.

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+* This document does not cover any requirements relating to where a Controller is additionally acting as a Node (e.g. receiving monitoring information via IS-07).
+
+## General
+
+### User
+Where this document refers to the "user" of a Controller, this includes both human operators who drive the Controller manually and automation systems that drive the Controller programmatically.
+
+### HTTP APIs
+#### Trailing Slashes
+APIs may advertise URLs with or without a trailing slash.
```
garethsb: My personal preference would be for more vertical whitespace in Markdown...

```suggestion
### HTTP APIs

#### Trailing Slashes

APIs may advertise URLs with or without a trailing slash.
```

garethsb: Please consider changing lowercase `may` to something that can't be confused with conformance language.

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+
+_(c) AMWA 2022, CC Attribution-NoDerivatives 4.0 International (CC BY-ND 4.0)_
+
+## Informative
+
+A Controller is Client software that interacts with the NMOS APIs to discover, connect and manage devices within a networked media system.
+
+* This document includes normative references to be followed when implementing a Controller.
+* This document covers how the Controller interacts with the NMOS APIs only.
+It does not cover other features of the Controller software, such as presentation.
```
garethsb: Should that have its own bullet `* ` or be indented with two spaces to be part of previous bullet?

docs/Controllers.md
```diff
@@ -0,0 +1,87 @@
...
+
+### HTTP APIs
+#### Trailing Slashes
+APIs may advertise URLs with or without a trailing slash.
+Controllers appending paths to `href` type attributes MUST support both cases, avoiding doubled or missing slashes.
+
+Controllers performing requests other than GET or HEAD (i.e PUT, POST, DELETE, OPTIONS etc.) MUST use URLs with no trailing slash present.
+
+#### API Version
+Controllers are responsible for identifying the correct API version they require.
```
garethsb: What does that actually mean?

docs/Controllers.md
garethsb: I see a double space before `Source`.

garethsb (commented):
Quick first review.

docs/Controllers.md
jonathan-r-thorpe: I believe the intent of the statement is to indicate to a user when Senders disappear from the registry for some reason and then reappear at some later point, but in a timely way. @garethsb @andrewbonney Do you have suggestions for better/ more consistent wording than 'online' and 'offline', (and for that matter dynamically) that might be being used in other NMOS specifications?

docs/Controllers.md
jonathan-r-thorpe: Noted - this should be broadened to include other values. It's worth pointing out that the General section of this Controllers document is intended to be common for all the Controller requirements documents across all affected ISs. As such this was intended to be boiler plate (that could be potentially factored into its own document). The clauses in the General section were lifted from existing requirements in IS-04, IS-05 and IS-08.  Therefore, adopting of a 'common' General section will necessarily mean repetition of requirements for certain specs.  The alternative is to tailor the General section to the given spec. This might be preferable if a common set of General requirements feels unworkable.

docs/Controllers.md
jonathan-r-thorpe: Noted - will check for use of conformance language. Also will check apostrophe use. Note that this sentence was lifted from [IS-08](https://specs.amwa.tv/is-08/releases/v1.0.1/docs/2.1._APIs_-_Client_Side_Implementation.html#failure-modes) for future updates to that spec.

docs/Controllers.md
jonathan-r-thorpe: Agreed this should be broadened. Note that this was sentence was lifted from [IS-08 ](https://specs.amwa.tv/is-08/releases/v1.0.1/docs/2.1._APIs_-_Client_Side_Implementation.html#failure-modes) for future updates to that spec.

docs/Controllers.md
jonathan-r-thorpe: Will check for consistent capitalization of Resource within the Controllers document.

docs/Controllers.md
jonathan-r-thorpe: Should be indented.

docs/Controllers.md
jonathan-r-thorpe: Noted and agreed. For reference this sentence was lifted from the [AMWA Wiki](https://github.com/AMWA-TV/nmos/wiki/Common-Issues#http-apis) for future updates.

docs/Controllers.md
jonathan-r-thorpe: Good question, given that it has lost the context from the [IS-05](https://specs.amwa.tv/is-05/releases/v1.1.1/docs/2.0._APIs.html#common-api-base-resource) specification where it originated, we should repeat the context here.

docs/Controllers.md
jonathan-r-thorpe: Noted.

docs/Controllers.md
jonathan-r-thorpe: See [this comment](https://github.com/AMWA-TV/is-04/pull/179/files#r809890875).

docs/Controllers.md
jonathan-r-thorpe: This perhaps lacks context. It is lifted from the [AMWA Wiki](https://github.com/AMWA-TV/nmos/wiki/IS-04-Client#best-practice), but that sentence links to the [Basic Queries section](https://github.com/AMWA-TV/is-04/blob/v1.2.x/docs/2.5.%20APIs%20-%20Query%20Parameters.md#basic-queries) of the [APIs: Query Parameters document](https://github.com/AMWA-TV/is-04/blob/v1.2.x/docs/2.5.%20APIs%20-%20Query%20Parameters.md#apis-query-parameters). The definition of Basic Queries there is more restrictive than you example, but perhaps not explicitly so. In that document it's referred to as a 'query filter' which is perhaps terminology we should adopt here, along with a link to that section of the specification.

docs/Controllers.md
jonathan-r-thorpe: This is sourced from the [AMWA Wiki](https://github.com/AMWA-TV/nmos/wiki/Generic-API-Client#websockets--subscriptions).  Is the problem the wording or the intent?

docs/Controllers.md
jonathan-r-thorpe: Noted on capitalization. Similar to the above [comment](https://github.com/AMWA-TV/is-04/pull/179/files#r810040878), this is sourced from the [AMWA Wiki](https://github.com/AMWA-TV/nmos/wiki/Generic-API-Client#websockets--subscriptions). It sounds like the intent is in doubt?

docs/Controllers.md
jonathan-r-thorpe: Lifted from the [AMWA Wiki](https://github.com/AMWA-TV/nmos/wiki/IS-04-Client#best-practice), but I think some context has been lost.  The Wiki says "If a WebSocket connection fails, do not attempt to reconnect to it as messages may have been missed". That should be included here as justification.

docs/Controllers.md
jonathan-r-thorpe: Noted on capitalization. Similar to this [comment](https://github.com/AMWA-TV/is-04/pull/179/files#r810040878), the justification would be lifted from the [AMWA Wiki](https://github.com/AMWA-TV/nmos/wiki/Generic-API-Client#websockets--subscriptions). Is this justification good enough?

docs/Controllers.md
jonathan-r-thorpe: Is the subscription mechanism not part of the Query API? How do we best identify the superset of HTTP and WebSocket interface?

docs/Controllers.md
garethsb: I checked JT-NM Tested test plan. Section 7.3 **Updating the list of resources dynamically from the IS-04
Registry** uses the phrases "putting offline/online" but explains what these mean (disconnecting from/reconnecting to the network). I quite like the description of the pass criteria:

> Changes in presence/absence of Reference Sender shall be reflected in the BCuT’s UI after a maximum of 30 seconds.

This wording makes clear that its about whether the Sender is currently present in the Registry or not, which is therefore dependent on the garbage collection interval, etc. (It also seems to me to permit that a Controller can simply stop showing the Sender at all when it's absent, which "indicate to the user" does perhaps not do so well?)

The [IS-04 Behaviour - Registration](https://specs.amwa.tv/is-04/releases/v1.3.1/docs/4.1._Behaviour_-_Registration.html#network-disconnect--reconnect) documentation also talks about registration and unregistration, disconnection and reconnection. This and other parts of the spec talk about adding and removing resources from the Registry.

* offline/online (and unavailable/available, inaccessible/accessible)
* disconnected/(re)connnected
* absent/present
 * removed/added
* unregistered/(re)registered

I like the later ones more than the earlier ones as they are more clearly about a Sender's registration status not its actual availability on the network or its stream status (which is obviously useful but I think not the point here?).

"Dynamically", I think, is shorthand for "without requiring user instigation and in a timely manner", right?

docs/Controllers.md
garethsb: Good point. Action item taken for the Architecture Review group to consider how to track these kind of comments.

docs/Controllers.md
garethsb: Good point. Action item taken for the Architecture Review group to consider how to track these kind of comments.

docs/Controllers.md
garethsb: 🤔

docs/Controllers.md
garethsb: Intent. I'm prepared to be persuaded, but AMWA Wiki has status lower than an INFO doc, so let's think about it critically.

docs/Controllers.md
garethsb: Same response.

docs/Controllers.md
garethsb: That justification is plain wrong as far as I'm aware. Reconnecting to a WebSocket results in all resources that match the `params` being sent again!

docs/Controllers.md
garethsb: Good questions, not sure.

docs/Controllers.md
jonathan-r-thorpe: Understood, although the Wiki recommendations do not appear to conflict with IS-04, just express preference for subscription over REST API. @peterbrightwell do you know the provenance of the 'advice' regarding a preference for subscriptions?

Also, while noting the lower status of the Wiki, the Wiki's NMOS Glossary *is* referenced by [IS-10](https://github.com/AMWA-TV/is-10/blob/019e0e3b4181addabbc2e526c79f44d99b09e20d/docs/Definitions.md) and [BCP-003-03](https://github.com/AMWA-TV/bcp-003-03/blob/7f72f432e65d578230adb800fc16d9ca1043795e/docs/Certificate%20Provisioning.md) giving it a normative air. However, also worth noting that this now links [to a page](https://github.com/AMWA-TV/nmos/wiki/Glossary), which in turn contains a [broken link](https://specs.amwa.tv/nmos/branches/main/docs/4.0._Glossary.html).

docs/Controllers.md
jonathan-r-thorpe: Assuming you mean that the advice on the Wiki is plain wrong (rather than my paraphrasing) I feel we might need the originator of the Wiki advice to make their case before we remove it from the PR (and by extension, the Wiki?). @peterbrightwell do you know where this advice originated?

docs/Controllers.md
jonathan-r-thorpe: The [Behaviour: Querying](https://github.com/AMWA-TV/is-04/blob/v1.3.x/docs/Behaviour%20-%20Querying.md) document says (my emphasis): 

> The Query API provides a read-only interface to the Nodes and their sub-resources which are currently held in the registry. Monitoring of these resources can be performed using the **HTTP API**, or using a **WebSocket connection** (RFC 6455), which is suggested for most applications.

So that appears to indicate that Query API is the superset, with HTTP API and WebSocket subscriptions as constituent parts.

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+* `MINOR` increments will be performed for non-breaking changes (such as the addition of attributes in a response)
+* `MAJOR` increments will be performed for breaking changes (such as the renaming of a resource or attribute)
+
+Versions MUST be represented as complete strings. Parsing MUST proceed as follows: separate into two strings, using the point (.) as a delimiter. Compare integer representations of `MAJOR`, `MINOR` version (such that v1.12 is greater than v1.5).
+
+Implementers of Controllers are RECOMMENDED to support multiple versions of the NMOS APIs simultaneously in order to ease the upgrade process in live facilities.
+
+#### API Common Keys
+
+Controllers SHOULD follow the requirements for common APi keys specified in the [IS-04 APIs: Common Keys](APIs%20-%20Common%20Keys.md) document including the requirements regarding [use of URNs](APIs%20-%20Common%20Keys.md#use-of-urns).
```
cristian-recoseanu:
```suggestion
Controllers SHOULD follow the requirements for common API keys specified in the [IS-04 APIs: Common Keys](APIs%20-%20Common%20Keys.md) document including the requirements regarding [use of URNs](APIs%20-%20Common%20Keys.md#use-of-urns).
```

cristian-recoseanu (changes requested):

docs/Controllers.md
peterbrightwell: The advice on the wiki was based on common sense at the time, as paging through a lot of results to find a small change for updating could be inefficient.  Maybe there are other ways for finding incremental updates that would be helpful but that would be require new investigations.

The Wiki shouldn't ever be referenced by the specs (so any links should be changed).  The Glossary is now at <https://specs.amwa.tv/nmos/branches/main/docs/Glossary.html>, and the Wiki glossary page directs the reader there.

docs/Controllers.md
peterbrightwell: Minor comment: we usually use `-` (rather than `*` for bullets in NMOS specs), although the linter allows both if they used consistently.

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+The NMOS APIs use HTTP status codes to indicate success, failure and other cases to Controllers as per [RFC 7231](https://tools.ietf.org/html/rfc7231) and related standards.
+
+As explicit handling of every possible HTTP response code is not expected, Controllers MUST implement generic handling for ranges of response codes (1xx, 2xx, 3xx, 4xx and 5xx).
+However, where the RAML specification of an API specifies explicit response codes the Controller SHOULD handle these cases explicitly.
+
+For Controllers performing `GET` and `HEAD` requests, using these methods SHOULD correctly handle a `301` (Moved Permanently) response.
+
+When a 301 is supported, the Controller MUST follow the redirect in order to retrieve the required response payload.
+
+If a Controller receives a HTTP 5xx or 4xx response code from the API, a failure has occurred.
```
peterbrightwell:
```suggestion
If a Controller receives an HTTP 5xx or 4xx response code from the API, a failure has occurred.
```

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+The Controller SHOULD display the content of the response's error field to the user if possible, and indicate that the resource is likely to be in an unknown state.
+The Controller SHOULD also refresh the endpoints of the relevant resources to ensure the Controller is accurately reflecting the current state of the API.
+
+## Registry Service Discovery	
+
+In order to locate the Registry, the Controller SHOULD support all of the following:
```
peterbrightwell: locate the Registration and Query Services?

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+
+If a WebSocket connection fails, then an attempt to reconnect to the WebSocket SHOULD NOT be attempted. Instead, a new subscription SHOULD be created with this API or a different one if required. If all available APIs return errors, an exponential backoff algorithm SHOULD be used when retrying until a success code is returned.
+
+## Dynamic Update of Resources
+The Controller MUST be capable of using the Registry's IS-04 Query API to discover and dynamically update the state of any registered resource, including Node, Device,  Source, Flow, Sender, and Receiver.
+
+* The Controller MUST indicate available Senders to the user.
+* The Controller MUST reflect changes in presence/absence of Senders to the user after a maximum of 30 seconds.
+* The Controller MUST indicate available Receivers to the user which have an IS-05 Connection API.
+  However, the Controller can choose not to display discovered Receivers without an IS-05 Connection API.
```
peterbrightwell: Should "can choose not to display" be "MAY omit display of"?  Or if not normative it could be  "might" rather than "can"?

docs/Controllers.md
peterbrightwell: This will have come from @andrewbonney some time ago.  I suggest a slack discussion with @garethsb to "sync" everyone's position :)

peterbrightwell: Have discussed with Andrew, and Gareth's info on this is likely more up-to-date, so probably best to remove from this PR (and Wiki)

garethsb: Added related issue https://github.com/AMWA-TV/nmos-testing/issues/664.

What are we going to do here then... Just delete this paragraph, or is there something else we can say?

docs/Controllers.md
garethsb: > Maybe there are other ways for finding incremental updates that would be helpful but that would be require new investigations.

😱 I don't think it requires new investigations. The whole paging mechanism is designed to make that easy!

1. The Query API defaults to `paging.order=update` on requests.
2. The `Link` response header provides the URL to request the next lot of updates (`rel="next"`).
3. That URL will use the `paging.since` query parameter with the response `X-Paging-Until` header value.

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+
+Controllers performing requests other than `GET` or `HEAD` (i.e `PUT`, `POST`, `DELETE`, `OPTIONS` etc.) MUST use URLs with no trailing slash present.
+
+#### API Versions
+
+The versioning format is `v<MAJOR>.<MINOR>`
+* `MINOR` increments will be performed for non-breaking changes (such as the addition of attributes in a response)
+* `MAJOR` increments will be performed for breaking changes (such as the renaming of a resource or attribute)
+
+Versions MUST be represented as complete strings. Parsing MUST proceed as follows: separate into two strings, using the point (.) as a delimiter. Compare integer representations of `MAJOR`, `MINOR` version (such that v1.12 is greater than v1.5).
```
garethsb:
```suggestion
Versions MUST be represented as complete strings. Parsing MUST proceed as follows: separate into two strings, using the point (`.`) as a delimiter. Compare integer representations of `MAJOR`, `MINOR` version (such that v1.12 is greater than v1.5).
```

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+
+#### API Common Keys
+
+Controllers SHOULD follow the requirements for common API keys specified in the [IS-04 APIs: Common Keys](APIs%20-%20Common%20Keys.md) document including the requirements regarding [use of URNs](APIs%20-%20Common%20Keys.md#use-of-urns).
+
+#### Error Codes & Responses
+
+The NMOS APIs use HTTP status codes to indicate success, failure and other cases to Controllers as per [RFC 7231](https://tools.ietf.org/html/rfc7231) and related standards.
+
+As explicit handling of every possible HTTP response code is not expected, Controllers MUST implement generic handling for ranges of response codes (1xx, 2xx, 3xx, 4xx and 5xx).
```
garethsb:
```suggestion
As explicit handling of every possible HTTP response code is not expected, Controllers MUST implement generic handling for ranges of response codes (`1xx`, `2xx`, `3xx`, `4xx` and `5xx`).
```

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+However, where the RAML specification of an API specifies explicit response codes the Controller SHOULD handle these cases explicitly.
+
+For Controllers performing `GET` and `HEAD` requests, using these methods SHOULD correctly handle a `301` (Moved Permanently) response.
+
+When a 301 is supported, the Controller MUST follow the redirect in order to retrieve the required response payload.
```
garethsb:
```suggestion
When a `301` is supported, the Controller MUST follow the redirect in order to retrieve the required response payload.
```

garethsb: This seems like strange conformance language: "SHOULD handle a `301` [and if it does] MUST follow the redirect"

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+
+If a Controller receives an HTTP 5xx or 4xx response code from the API, a failure has occurred.
```
garethsb:
```suggestion
If a Controller receives an HTTP `5xx` or `4xx` response code from the API, a failure has occurred.
```

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+The Controller SHOULD display the content of the response's error field to the user if possible, and indicate that the resource is likely to be in an unknown state.
+The Controller SHOULD also refresh the endpoints of the relevant resources to ensure the Controller is accurately reflecting the current state of the API.
```
garethsb: @jonathan-r-thorpe I think a resource is possibly only in an unknown state after a `5xx` response. (I notice that originally the PR only mentioned `500` up until https://github.com/AMWA-TV/is-04/commit/c77380b93101d39e6c12926ca8a664416bd0eeca#diff-b799a84e93704a0eb1c1d6fe13ea49c46bd08755d9c47ef3e7aa2073eefb837eR57.)

A `4xx` response indicates the API has determined a problem with the request, so I think the resource SHOULD be left unchanged in that case? @pedro-alves-ferreira agreed in weekly ARG call.

However, conversely, I don't actually see anything in https://datatracker.ietf.org/doc/html/rfc7231#section-6.6 that explicitly says resources could be in an unknown state after a `5xx` error. 🤔

Moreover, the guidance about refreshing the relevant resources is perhaps too broad, e.g. after a `400` (Bad Request) response or a `501` (Not Implemented) response, a Client SHOULD NOT make the same request (`/bang-head-against-wall`).

docs/Controllers.md
garethsb: In the rest of IS-04, I thought it normally talks about discovering the Registration API or the Query API.
In the case of Controllers, their acting as a Node is explicitly not covered, so I think it's only the Query API that this section needs to be concerned with?

```suggestion
In order to locate the Query API, the Controller SHOULD support all of the following:
```

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+
+If a WebSocket connection fails, then an attempt to reconnect to the WebSocket SHOULD NOT be attempted. Instead, a new subscription SHOULD be created with this API or a different one if required. If all available APIs return errors, an exponential backoff algorithm SHOULD be used when retrying until a success code is returned.
+
+## Dynamic Update of Resources
+The Controller MUST be capable of using the Registry's IS-04 Query API to discover and dynamically update the state of any registered resource, including Node, Device,  Source, Flow, Sender, and Receiver.
+
+* The Controller MUST indicate available Senders to the user.
+* The Controller MUST reflect changes in presence/absence of Senders to the user after a maximum of 30 seconds.
+* The Controller MUST indicate available Receivers to the user which have an IS-05 Connection API.
+  However, the Controller can choose not to display discovered Receivers without an IS-05 Connection API.
```
garethsb: I'm asking myself why the language is different for Senders and Receivers?
Do we REQUIRE Controllers to be able to connect (IS-05 supporting) Receivers to IS-04 only Senders?

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+If using the RESTful Query API rather than WebSockets, Pagination requirements MUST be implemented as specified in the [Pagination section of the APIs: Query Parameters document](APIs%20-%20Query%20Parameters.md#pagination) in this specification.
+	
+## WebSockets & Subscriptions	
+
+Where a WebSocket or other subscription based mechanism is provided for Controller usage, it is strongly RECOMMENDED that Controllers make use of this and do not use the API resources directly.
+
+If a WebSocket connection fails, then an attempt to reconnect to the WebSocket SHOULD NOT be attempted. Instead, a new subscription SHOULD be created with this API or a different one if required. If all available APIs return errors, an exponential backoff algorithm SHOULD be used when retrying until a success code is returned.
+
+## Dynamic Update of Resources
+The Controller MUST be capable of using the Registry's IS-04 Query API to discover and dynamically update the state of any registered resource, including Node, Device,  Source, Flow, Sender, and Receiver.
```
garethsb:
```suggestion
The Controller MUST be capable of using the Query API to discover and dynamically update the state of any registered resource, including Node, Device, Source, Flow, Sender, and Receiver.
```

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+In large systems, API resources rely upon pagination in order to return high volumes of data.
+This can be both hard to keep track of for a client, and require a large number of requests in order to scan the entire data set.
+
+For this reason it is RECOMMENDED to use the RESTful QUERY API for debug and development purposes only.
+
+If using the RESTful Query API rather than WebSockets, Pagination requirements MUST be implemented as specified in the [Pagination section of the APIs: Query Parameters document](APIs%20-%20Query%20Parameters.md#pagination) in this specification.
+	
+## WebSockets & Subscriptions	
+
+Where a WebSocket or other subscription based mechanism is provided for Controller usage, it is strongly RECOMMENDED that Controllers make use of this and do not use the API resources directly.
```
garethsb: This sounds like it was written to be used in a document covering several APIs with different subscription mechanisms? Do we have any others? Here anyway I suggest:

```suggestion
It is RECOMMENDED that Controllers make use of the WebSocket subscription mechanism and do not use the HTTP API resources directly.
```

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+The Controller SHOULD offer unicast DNS-SD as the default mechanism. 
+
+Controllers SHOULD observe and interpret all of the TXT records returned with the DNS Service Discovery responses according to the requirements for Query API Clients specified in the [API Paths section of the APIs document](APIs.md#api-paths) in this specification.
+
+## Query API
+
+The Controller MUST be capable of using the Registry's IS-04 Query API to discover any registered resource, including Node, Device,  Source, Flow, Sender, and Receiver,
+as described in the [APIs section](README.md#apis) of this specification.
+
+The Controller MUST use the Registry's IS-04 Query API either via the HTTP API or by requesting WebSocket subscriptions.
```
garethsb:
```suggestion
The Controller MUST use the Query API either via the HTTP API or by requesting WebSocket subscriptions.
```

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+* Multicast DNS Service Discovery (mDNS).
+* Direct configuration with the location of a preferred Registry.
+
+The Controller SHOULD offer unicast DNS-SD as the default mechanism. 
+
+Controllers SHOULD observe and interpret all of the TXT records returned with the DNS Service Discovery responses according to the requirements for Query API Clients specified in the [API Paths section of the APIs document](APIs.md#api-paths) in this specification.
+
+## Query API
+
+The Controller MUST be capable of using the Registry's IS-04 Query API to discover any registered resource, including Node, Device,  Source, Flow, Sender, and Receiver,
```
garethsb:
```suggestion
The Controller MUST be capable of using the Query API to discover any registered resource, including Node, Device, Source, Flow, Sender, and Receiver,
```

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+If a Controller receives an HTTP 5xx or 4xx response code from the API, a failure has occurred.
+The Controller SHOULD display the content of the response's error field to the user if possible, and indicate that the resource is likely to be in an unknown state.
+The Controller SHOULD also refresh the endpoints of the relevant resources to ensure the Controller is accurately reflecting the current state of the API.
+
+## Registry Service Discovery	
+
+In order to locate the Registry, the Controller SHOULD support all of the following: 
+* Unicast DNS Service Discovery (DNS-SD).
+* Multicast DNS Service Discovery (mDNS).
+* Direct configuration with the location of a preferred Registry.
```
garethsb:
```suggestion
* Direct configuration with the location of a preferred Query API.
```

docs/Controllers.md
garethsb:
```suggestion
  However, the Controller MAY choose to omit discovered Receivers without an IS-05 Connection API.
```

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+If a Controller receives an HTTP `5xx` or `4xx` response code from the API, a failure has occurred.
+The Controller SHOULD display the content of the response's error field to the user if possible, and indicate that the resource is likely to be in an unknown state.
+The Controller SHOULD also refresh the endpoints of the relevant resources to ensure the Controller is accurately reflecting the current state of the API.
+
+## Registry Service Discovery	
+
+In order to locate the Query API, the Controller SHOULD support all of the following:
+* Unicast DNS Service Discovery (DNS-SD).
+* Multicast DNS Service Discovery (mDNS).
+* Direct configuration with the location of a preferred Query API.
```
garethsb:
```suggestion
## Query API Discovery	

NMOS [Discovery](Discovery.md) makes use of the DNS Service Discovery (DNS-SD) protocol. The [Query API: Client Interaction Procedure](Discovery%20-%20Registered%20Operation.md#client-interaction-procedure-1) section of this specification describes the requirements in detail.

In order to locate the Query API, the Controller SHOULD support all of the following:

* Unicast DNS
* Multicast DNS (mDNS)
* Direct configuration with the location of a preferred Query API
```

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+
+The Controller SHOULD offer unicast DNS-SD as the default mechanism. 
+
+Controllers SHOULD observe and interpret all of the TXT records returned with the DNS Service Discovery responses according to the requirements for Query API Clients specified in the [API Paths section of the APIs document](APIs.md#api-paths) in this specification.
+
+## Query API
+
+The Controller MUST be capable of using the Query API to discover any registered resource, including Node, Device, Source, Flow, Sender, and Receiver,
+as described in the [APIs section](README.md#apis) of this specification.
```
garethsb: Hmm, I don't think links to _README.md_ will work as it isn't rendered in the same way as other _docs/_.

```suggestion
as described in the [APIs](APIs.md) section of this specification.
```

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+
+The Controller MUST use the Query API either via the HTTP API or by requesting WebSocket subscriptions.
+
+When using the Query API, query filters SHOULD be used (and advanced query language where available) to cut down on the volume of resources returned to the Controller, as specified in the [APIs: Query Parameters](APIs%20-%20Query%20Parameters.md) document.
+
+Controllers SHOULD adhere to the version downgrade requirements for Query API Clients specified in the [Requirements for Query API Clients section of the Upgrade Path document](Upgrade%20Path.md#requirements-for-query-api-clients) in this specification.
```
garethsb:
```suggestion
Controllers SHOULD adhere to the version downgrade requirements for Query API Clients specified in the [Upgrade Path: Requirements for Query API Clients](Upgrade%20Path.md#requirements-for-query-api-clients) section of this specification.
```

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+## Registry Service Discovery	
+
+In order to locate the Query API, the Controller SHOULD support all of the following:
+* Unicast DNS Service Discovery (DNS-SD).
+* Multicast DNS Service Discovery (mDNS).
+* Direct configuration with the location of a preferred Query API.
+
+The Controller SHOULD offer unicast DNS-SD as the default mechanism. 
+
+Controllers SHOULD observe and interpret all of the TXT records returned with the DNS Service Discovery responses according to the requirements for Query API Clients specified in the [API Paths section of the APIs document](APIs.md#api-paths) in this specification.
```
garethsb:
```suggestion
Controllers SHOULD observe and interpret all of the TXT records returned with the DNS Service Discovery responses according to the requirements for Query API Clients specified in the [APIs: API Paths](APIs.md#api-paths) section of this specification.
```

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+Controllers SHOULD adhere to the version downgrade requirements for Query API Clients specified in the [Requirements for Query API Clients section of the Upgrade Path document](Upgrade%20Path.md#requirements-for-query-api-clients) in this specification.
+
+## Pagination
+
+In large systems, API resources rely upon pagination in order to return high volumes of data.
+This can be both hard to keep track of for a client, and require a large number of requests in order to scan the entire data set.
+
+For this reason it is RECOMMENDED to use the RESTful QUERY API for debug and development purposes only.
+
+If using the HTTP API rather than WebSocket subscriptions, pagination requirements MUST be implemented as specified in the [Pagination section of the APIs: Query Parameters document](APIs%20-%20Query%20Parameters.md#pagination) in this specification.
```
garethsb:
```suggestion
If using the HTTP API rather than WebSocket subscriptions, pagination requirements MUST be implemented as specified in the [APIs: Query Parameters: Pagination](APIs%20-%20Query%20Parameters.md#pagination) section of this specification.
```

garethsb (commented):
Fix and shorten links

docs/README.md
```diff
@@ -24,6 +24,9 @@
   - [Querying](Behaviour%20-%20Querying.md)
   - [Nodes](Behaviour%20-%20Nodes.md)
 
+### Controllers
+- [General](Controllers.md)
```
garethsb:
```suggestion
### Controllers

- [General](Controllers.md)
```

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+
+* This document includes normative references to be followed when implementing a Controller.
+* This document covers how the Controller interacts with the NMOS APIs only.
+  It does not cover other features of the Controller software, such as presentation.
+* This document does not cover any requirements relating to where a Controller is additionally acting as a Node (e.g. receiving monitoring information via IS-07).
+
+## General
+
+### User
+Where this document refers to the "user" of a Controller, this includes both human operators who drive the Controller manually and automation systems that drive the Controller programmatically.
```
garethsb:
```suggestion
### User

Where this document refers to the "user" of a Controller, this includes both human operators who drive the Controller manually and automation systems that drive the Controller programmatically.
```

docs/Controllers.md
```diff
@@ -0,0 +1,99 @@
...
+If using the HTTP API rather than WebSocket subscriptions, pagination requirements MUST be implemented as specified in the [APIs: Query Parameters: Pagination](APIs%20-%20Query%20Parameters.md#pagination) section of this specification.
+	
+## WebSockets & Subscriptions	
+
+Where a WebSocket or other subscription based mechanism is provided for Controller usage, it is strongly RECOMMENDED that Controllers make use of this and do not use the API resources directly.
+
+If a WebSocket connection fails, then an attempt to reconnect to the WebSocket SHOULD NOT be attempted. Instead, a new subscription SHOULD be created with this API or a different one if required. If all available APIs return errors, an exponential backoff algorithm SHOULD be used when retrying until a success code is returned.
+
+## Dynamic Update of Resources
+The Controller MUST be capable of using the Query API to discover and dynamically update the state of any registered resource, including Node, Device, Source, Flow, Sender, and Receiver.
```
garethsb:
```suggestion
## Dynamic Update of Resources

The Controller MUST be capable of using the Query API to discover and dynamically update the state of any registered resource, including Node, Device, Source, Flow, Sender, and Receiver.
```

garethsb (commented):
Whitespace

docs/Controllers.md
```diff
@@ -0,0 +1,104 @@
...
+## Pagination
+
+In large systems, API resources rely upon pagination in order to return high volumes of data.
+This can be both hard to keep track of for a client, and require a large number of requests in order to scan the entire data set.
+
+For this reason it is RECOMMENDED to use the RESTful QUERY API for debug and development purposes only.
+
+If using the HTTP API rather than WebSocket subscriptions, pagination requirements MUST be implemented as specified in the [APIs: Query Parameters: Pagination](APIs%20-%20Query%20Parameters.md#pagination) section of this specification.
+	
+## WebSockets & Subscriptions
```
garethsb:
```suggestion
## WebSocket Subscriptions	
```

docs/Controllers.md
garethsb: https://github.com/AMWA-TV/is-04/blob/v1.2.x/docs/4.2.%20Behaviour%20-%20Querying.md#websocket-protocol says:

> Using data Grains, the websocket protocol provides a client with the current state of the contents of the registry at the time of _subscription (my emphasis)_. It subsequently notifies clients of any changes.

I'm 99% certain that was intended to mean "at the time of _connection_" but I'd like @andrewbonney's opinion...

docs/Controllers.md
garethsb: Agreed. During review let's ensure consistent usage of "HTTP API" where that is specifically intended.

docs/Controllers.md
garethsb: Seems to have been deleted now.

docs/Controllers.md
garethsb: Current statement is now:
> * The Controller MUST reflect changes in presence/absence of Senders to the user after a maximum of 30 seconds.

So I think this has been dealt with... There's a comment https://github.com/AMWA-TV/is-04/pull/179#discussion_r848087550 to discuss why this language isn't the same for both Senders and Receivers.

docs/Controllers.md
garethsb: Ah - just noticed I was on the v1.2.x docs there... sure enough the language is corrected in v1.3.x - https://github.com/AMWA-TV/is-04/blob/v1.3.x/docs/Behaviour%20-%20Querying.md#websocket-messages says:

> The server provides a client with the current state of the contents of the registry at the time of their WebSocket connection, using messages which are known as data Grains. It subsequently notifies clients of any changes via additional data Grain messages.

garethsb: Ah - just noticed I was on the v1.2.x docs there... sure enough the language is corrected in v1.3.x - https://github.com/AMWA-TV/is-04/blob/v1.3.x/docs/Behaviour%20-%20Querying.md#websocket-messages says:

> The server provides a client with the current state of the contents of the registry at the time of their WebSocket connection, using messages which are known as data Grains. It subsequently notifies clients of any changes via additional data Grain messages.

docs/Controllers.md
```diff
@@ -0,0 +1,104 @@
...
+The Controller SHOULD offer unicast DNS-SD as the default mechanism. 
+
+Controllers SHOULD observe and interpret all of the TXT records returned with the DNS Service Discovery responses according to the requirements for Query API Clients specified in the [APIs: API Paths](APIs.md#api-paths) section of this specification.
+
+## Query API
+
+The Controller MUST be capable of using the Query API to discover any registered resource, including Node, Device, Source, Flow, Sender, and Receiver,
+as described in the [APIs](APIs.md) section of this specification.
+
+The Controller MUST use the Query API either via the HTTP API or by requesting WebSocket subscriptions.
```
garethsb: A ∪ A' = U 🤔 

The Query API is only accessible by one or other of those techniques, there isn't a third choice... What is this saying, and if it is useful at all, does it REQUIRE a conformance language key word?

docs/Controllers.md
garethsb: How about this?

```suggestion
The Controller SHOULD display the content of the response's `error` field to the user if possible.
The Controller SHOULD continue indicating this error status for the resource until it has been refreshed.
```
(or every resource covered by the same API?)

However, I'm not sure what to tell a Controller to do, regarding making further requests.

Do we need something similar to the guidance for Nodes' use of the Registration API?

> A `400` (Bad Request) error indicates a client error which is likely to be the result of a validation failure identified by the Registration API. The same request MUST NOT be re-attempted without corrective action being taken first.
...
> If a `5xx` error is encountered when interacting with all discoverable Registration APIs, clients SHOULD implement an exponential backoff algorithm in their next attempts until a non-`5xx` response code is received.

docs/Controllers.md
jonathan-r-thorpe: Resolving conversation as now outdated.

docs/Controllers.md
jonathan-r-thorpe: Resolving conversation as now outdated

docs/Controllers.md
jonathan-r-thorpe: Resolving conversation as now outdated

docs/Controllers.md
jonathan-r-thorpe: Resolving conversation as now outdated

docs/Controllers.md
jonathan-r-thorpe: Resolving conversation as now outdated

docs/Controllers.md
jonathan-r-thorpe: Resolving conversation as now outdated

docs/Controllers.md
jonathan-r-thorpe: Has this been discussed within the ARG? I would agree that pagination is intended to make it easier for a client rather than "hard to keep track of".  Shall we remove/edit this section?

docs/Controllers.md
jonathan-r-thorpe: OK, is the conclusion here that we remove line 79?

docs/Controllers.md
jonathan-r-thorpe: Regarding the provenance of this requirement, this was lifted from [IS-08](https://specs.amwa.tv/is-08/releases/v1.0.1/docs/2.1._APIs_-_Client_Side_Implementation.html#failure-modes).

jonathan-r-thorpe: In that context it requires refreshing of `map/active` and `map/activations` endpoints and was generalized here. 
But anyway, yes, adding guidance on handling errors would be useful I think.

docs/Controllers.md
jonathan-r-thorpe: Agreed 😊 This was originally from the JT-NM-Tested catalog, but I think something got lost in translation between there and here.

docs/Controllers.md
```diff
@@ -0,0 +1,95 @@
...
+
+The NMOS APIs use HTTP status codes to indicate success, failure and other cases to Controllers as per [RFC 7231](https://tools.ietf.org/html/rfc7231) and related standards.
+
+As explicit handling of every possible HTTP response code is not expected, Controllers MUST implement generic handling for ranges of response codes (`1xx`, `2xx`, `3xx`, `4xx` and `5xx`).
+However, where the RAML specification of an API specifies explicit response codes the Controller SHOULD handle these cases explicitly.
+
+For Controllers performing `GET` and `HEAD` requests, using these methods SHOULD correctly handle a `301` (Moved Permanently) response.
+
+When a `301` is supported, the Controller MUST follow the redirect in order to retrieve the required response payload.
+
```
jonathan-r-thorpe:
```suggestion
Controllers performing `GET` and `HEAD` requests SHOULD handle a `301` (Moved Permanently) response by following the redirect.
```
This is intended to side step the confusion caused by a SHOULD followed by a dependent MUST. How does it look?

docs/Controllers.md
garethsb: Resolved.

docs/Controllers.md
```diff
@@ -0,0 +1,96 @@
...
+#### Error Codes & Responses
+
+The NMOS APIs use HTTP status codes to indicate success, failure and other cases to Controllers as per [RFC 7231](https://tools.ietf.org/html/rfc7231) and related standards.
+
+As explicit handling of every possible HTTP response code is not expected, Controllers MUST implement generic handling for ranges of response codes (`1xx`, `2xx`, `3xx`, `4xx` and `5xx`).
+However, where the RAML specification of an API specifies explicit response codes the Controller SHOULD handle these cases explicitly.
+
+For Controllers performing `GET` and `HEAD` requests, using these methods SHOULD correctly handle a `301` (Moved Permanently) response.
+
+When a `301` is supported, the Controller MUST follow the redirect in order to retrieve the required response payload.
```
garethsb: After ARG review and discussion with @jonathan-r-thorpe, we propose making this a REQUIREMENT for Controllers rather than a RECOMMENDATION.

```suggestion
When a Controller performs `GET` and `HEAD` requests, it MUST correctly handle a `301` (Moved Permanently) response by following the redirect `Location` header.
```

docs/Controllers.md
```diff
@@ -0,0 +1,96 @@
...
+
+If a Controller receives an HTTP `5xx` or `4xx` response code from the API, a failure has occurred.
+The Controller SHOULD display the content of the response's `error` field to the user if possible.
+The Controller SHOULD continue indicating this error status for the resource until it has been refreshed.
```
garethsb: After ARG review, we propose keeping the RECOMMENDATION to display `error` messages, but omitting that for persistence or refresh as there are many different models of Server-Controller-User interaction which the Controller might wish to provide.

```suggestion
```

docs/Controllers.md
```diff
@@ -0,0 +1,96 @@
...
+
+In order to locate the Query API, the Controller SHOULD support all of the following:
+
+* Unicast DNS
+* Multicast DNS (mDNS)
+* Direct configuration with the location of a preferred Query API
+
+The Controller SHOULD offer unicast DNS-SD as the default mechanism. 
+
+Controllers SHOULD observe and interpret all of the TXT records returned with the DNS Service Discovery responses according to the requirements for Query API Clients specified in the [APIs: API Paths](APIs.md#api-paths) section of this specification.
```
garethsb: @garethsb suggestion, @jonathan-r-thorpe and @AMWA-TV/nmos-architecture-review to approve:

```suggestion
Controllers MUST observe and interpret all of the TXT records returned with the DNS Service Discovery responses according to the requirements for Query API Clients specified in the [APIs: API Paths](APIs.md#api-paths) section of this specification.
```

docs/Controllers.md
```diff
@@ -0,0 +1,96 @@
...
+* `MINOR` increments will be performed for non-breaking changes (such as the addition of attributes in a response)
+* `MAJOR` increments will be performed for breaking changes (such as the renaming of a resource or attribute)
+
+Versions MUST be represented as complete strings. Parsing MUST proceed as follows: separate into two strings, using the point (`.`) as a delimiter. Compare integer representations of `MAJOR`, `MINOR` version (such that v1.12 is greater than v1.5).
+
+Implementers of Controllers are RECOMMENDED to support multiple versions of the NMOS APIs simultaneously in order to ease the upgrade process in live facilities.
+
+#### API Common Keys
+
+Controllers SHOULD follow the requirements for common API keys specified in the [IS-04 APIs: Common Keys](APIs%20-%20Common%20Keys.md) document including the requirements regarding [use of URNs](APIs%20-%20Common%20Keys.md#use-of-urns).
```
garethsb: After discussion with ARG and @jonathan-r-thorpe, we propose to make this a REQUIREMENT since there is currently only one mention of clients in the referenced document and it is a REQUIREMENT.

```suggestion
Controllers MUST follow the requirements for common API keys specified in the [IS-04 APIs: Common Keys](APIs%20-%20Common%20Keys.md) document including the requirements regarding [use of URNs](APIs%20-%20Common%20Keys.md#use-of-urns).
```

docs/Controllers.md
```diff
@@ -0,0 +1,96 @@
...
+Controllers SHOULD observe and interpret all of the TXT records returned with the DNS Service Discovery responses according to the requirements for Query API Clients specified in the [APIs: API Paths](APIs.md#api-paths) section of this specification.
+
+## Query API
+
+The Controller MUST be capable of using the Query API to discover any registered resource, including Node, Device, Source, Flow, Sender, and Receiver,
+as described in the [APIs](APIs.md) section of this specification.
+
+When using the Query API, query filters SHOULD be used (and advanced query language where available) to cut down on the volume of resources returned to the Controller, as specified in the [APIs: Query Parameters](APIs%20-%20Query%20Parameters.md) document.
+
+Controllers SHOULD adhere to the version downgrade requirements for Query API Clients specified in the [Upgrade Path: Requirements for Query API Clients](Upgrade%20Path.md#requirements-for-query-api-clients) section of this specification.
```
garethsb: @jonathan-r-thorpe - I wonder if this ought to be REQUIRED as well? The referenced document section uses the term "query downgrade mechanism" rather than "version downgrade" however, so please would you confirm the intention of the drafting group?

jonathan-r-thorpe: This language was lifted from the [AMWA Wiki](https://github.com/AMWA-TV/nmos/wiki/IS-04-Client) which states:

> "Make use of the [Query API downgrade](https://specs.amwa.tv/is-04/branches/v1.3.x/docs/APIs_-_Query_Parameters.html#downgrade-queries) function wherever possible. This ensures that you can display Nodes and their resources via your interface, even if they only support a lower version of the API specifications."

but looking at the referenced clauses in the [Upgrade Path](https://specs.amwa.tv/is-04/releases/v1.3.1/docs/6.0._Upgrade_Path.html#requirements-for-registries-registration-and-query-apis) document the requirement are for Query API providers, not clients.  Perhaps the reference should be to  [Downgrade Queries](https://specs.amwa.tv/is-04/branches/v1.3.x/docs/APIs_-_Query_Parameters.html#downgrade-queries) instead?

docs/Controllers.md
jonathan-r-thorpe: LGTM

docs/Controllers.md
jonathan-r-thorpe: LGTM

docs/Controllers.md
jonathan-r-thorpe: LGTM

docs/Controllers.md
garethsb: Resolved here, but ARG should look at IS-08 at some point...

docs/Controllers.md
garethsb: The referenced section has one (pointless) MUST and then a number of SHOULD/SHOULD NOT/RECOMMENDED. Do you intend to change the force of any of these for Controllers via this reference?

jonathan-r-thorpe: So the intention was to reference the requirements *as is* from the referenced document, i.e. to the same force.

docs/Controllers.md
```diff
@@ -0,0 +1,90 @@
...
+Controllers MUST observe and interpret all of the TXT records returned with the DNS Service Discovery responses according to the requirements for Query API Clients specified in the [APIs: API Paths](APIs.md#api-paths) section of this specification.
+
+## Query API
+
+The Controller MUST be capable of using the Query API to discover any registered resource, including Node, Device, Source, Flow, Sender, and Receiver,
+as described in the [APIs](APIs.md) section of this specification.
+
+When using the Query API, query filters SHOULD be used (and advanced query language where available) to cut down on the volume of resources returned to the Controller, as specified in the [APIs: Query Parameters](APIs%20-%20Query%20Parameters.md) document.
+
+Controllers SHOULD adhere to the version downgrade requirements for Query API Clients specified in the [Upgrade Path: Requirements for Query API Clients](Upgrade%20Path.md#requirements-for-query-api-clients) section of this specification.
```
jonathan-r-thorpe:
```suggestion
Controllers MUST adhere to the [Requirements for Query API Clients](Upgrade%20Path.md#requirements-for-query-api-clients) found in this specification.
```
Maintaining the force of the referenced normative language.

docs/Controllers.md
```diff
@@ -0,0 +1,90 @@
...
+
+If using the HTTP API rather than WebSocket subscriptions, pagination requirements MUST be implemented as specified in the [APIs: Query Parameters: Pagination](APIs%20-%20Query%20Parameters.md#pagination) section of this specification.
+	
+## WebSocket Subscriptions	
+
+In order to avoid polling of the HTTP API, it is RECOMMENDED that Controllers use the WebSocket subscription mechanism.
+
+## Dynamic Update of Resources
+
+The Controller MUST be capable of using the Query API to discover and dynamically update the state of any registered resource, including Node, Device, Source, Flow, Sender, and Receiver.
```
jonathan-r-thorpe:
```suggestion
The Controller MUST be capable of using the Query API to discover and dynamically report the state of any registered resource, including Node, Device, Source, Flow, Sender, and Receiver.
```

docs/Controllers.md
garethsb: Currently has been resolved with MUST to ensure a Controller isn't going to be tripped up by redirects. This might be a case where existing software could be conformant to the existing Client requirements but not to the Controller reqs... but seriously, if you can't follow an  HTTP redirect for a `GET` request, and Node and Registry APIs are allowed to return them, we have an interop problem waiting to happen.

garethsb (approved):
LGTM :alien: :mage_man:

docs/Controllers.md
```diff
@@ -0,0 +1,90 @@
...
+Controllers MUST follow the requirements for common API keys specified in the [IS-04 APIs: Common Keys](APIs%20-%20Common%20Keys.md) document including the requirements regarding [use of URNs](APIs%20-%20Common%20Keys.md#use-of-urns).
+
+#### Error Codes & Responses
+
+The NMOS APIs use HTTP status codes to indicate success, failure and other cases to Controllers as per [RFC 7231](https://tools.ietf.org/html/rfc7231) and related standards.
+
+As explicit handling of every possible HTTP response code is not expected, Controllers MUST implement generic handling for ranges of response codes (`1xx`, `2xx`, `3xx`, `4xx` and `5xx`).
+However, where the API specifies explicit response codes the Controller SHOULD handle these cases explicitly.
+
+When a Controller performs `GET` and `HEAD` requests, it MUST correctly handle a `301` (Moved Permanently) response by following the redirect `Location` header.
```
garethsb: @cristian-recoseanu, @jonathan-r-thorpe and I discussed default and permitted redirect handling in typical software libraries, e.g. .NET Core, JavaScript Fetch API and Python Requests. The default behaviour is often to automatically follow redirects up to some limit except for e.g. a redirect from HTTPS to insecure HTTP.

That makes I wonder if we should modify this to reference that it's only to handle the fact Servers are permitted to redirect from the path with a trailing slash to the one without or vice-versa.

E.g.

```suggestion
When a Controller performs `GET` and `HEAD` requests, it MUST correctly handle a `301` (Moved Permanently) redirect to accommodate the API implementations permitted in the [APIs: URLs: Approach to Trailing Slashes](APIs.md#urls-approach-to-trailing-slashes) section of this specification.
```

jonathan-r-thorpe: Yes, that's a useful clarification for its intended use.

garethsb: :+1: It allows Controllers to follow best practice and refuse to follow insecure redirects, loops, etc. automatically

garethsb:
Approved and squashed for final review as #181. Congratulations and thank you, @jonathan-r-thorpe and the Controller Specification activity group.

Issue #176, opened 2021-08-26
Can Sender exist without a Flow?


N-Nagorny:
[Starting from v1.1](https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.3.x/docs/6.0.%20Upgrade%20Path.md#affected-keys-from-v11) `flow_id` can be `null` but Data Model [shows](https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.3.x/docs/images/class.png) that Sender always have one Flow. Could anyone answer whether Sender without an associated Flow is a valid case?

garethsb:
See https://github.com/AMWA-TV/nmos-discovery-registration/pull/4.

> Permit ʼnullʼ flow_id against a Sender in order to allow advertisement before internal device routing has occurred

So I think it's just that the diagram didn't get updated.

prince-chrismc:
I was not able to find any text or example... the only spot is the change log... perhaps this feature needs to be better document? Several controller require a flow/source to work.

garethsb:
Agreed, the Upgrade Path mention and the schema itself allowing `null` aren't really enough. It'd certainly make sense to (a) fix the Data Model diagram, (b) add some words like those from the original PR to the `description` at https://github.com/AMWA-TV/nmos-discovery-registration/blob/v1.3.x/APIs/schemas/sender.json#L26, and (c) to add a test for this in the Client Testing effort (👀 @jonathan-r-thorpe)

PR #182, opened 2022-05-12
Data model now draw.io based

peterbrightwell:
Data model now draw.io based :)


pedro-alves-ferreira (approved):
Nice. Was there any reason for mirroring it, so that the outputs are now on the left?

garethsb:
What's the difference between the Node-Device relationship and the Device-{Source,Flow,Sender,Receiver} ones? These both show composition (filled diamond) but  former doesn't have a direction arrowhead but the latter do. (And the IS-04 resources model the relationship in the opposite direction to the indicated arrowheads.)

Did we ought to indicate somehow that Grain and Connection are not modelled by IS-04 API resources?

The Source-Flow and Sender-Flow relationships are both shown as aggregation (unfilled diamond) but in the IS-04 resource model the relationships are modelled in opposite directions (Sender has `flow_id` and Flow has `source_id`).

The relationship between Sender and Receiver is shown as a simple association but the multiplicities have been reversed in the new diagram which doesn't look right, but I'm not sure it was correct before either. Do we want to show a Sender can send to 0..* Receivers, and a Receiver can receive from 0..1 Senders? (Or even, should we show many-many multicast as an option (many Senders sending to the same Receiver as well as many Receivers receiving from one Sender) which would make it 0..* at both ends?) Does it matter that the IS-04 resource properties `subscription.receiver_id` and `subscription.sender_id` _can't_ model both ends of the multicast cases?

garethsb:
Cf. https://github.com/AMWA-TV/is-11/blob/v1.0-dev/docs/images/is-11-data-model.png which is also a diagrams.net PNG.

It would be nice if the two diagrams ended up identical apart from the introduction of Input and Output in the IS-11 one.

alabou:
A relationship has changed:
was Sender 1 -> 0..* Receiver and is now Sender 0..* -> 1 Receiver
I think that the former was the right one.

cristian-recoseanu:
In the old diagram the arrowhead from a flow was pointing towards the source whereas in the new one the direction is reversed which might be wrong?

garethsb:
This PR needs to address the comment about the diagram in #176.

PR #183, opened 2022-05-17
Update Lint to include pull requests

garethsb:


garethsb:
Decided in @AMWA-TV/nmos-architecture-review discussion to lint all branches as well as PRs, to ensure potential contributions get early feedback. See https://github.com/AMWA-TV/is-template/pull/48.

PR #186, opened 2022-06-22
Clarify meaning of Sender flow_id being null

### LINKED ISSUE #176: Can Sender exist without a Flow?

garethsb:
With #182, resolves #176.

### DIFF
--- a/APIs/schemas/sender.json
+++ b/APIs/schemas/sender.json
@@ -19,11 +19,11 @@
         "caps": {
           "description": "Capabilities of this sender",
           "type": "object",
-          "properties":{
+          "properties": {
           }
         },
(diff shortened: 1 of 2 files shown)


peterbrightwell (approved):
Looks good

PR #187, opened 2022-06-22
Directly refer to the NMOS Parameter Registers

garethsb:


garethsb:
Is there anywhere else in IS-04 that we should now reference the NMOS Parameter Registers ?

peterbrightwell (approved):
Looks good

PR #188, opened 2022-06-22
Clarify error handling behaviour applies to both registration and heartbeat

### LINKED ISSUE #169: HTTP 500 Error Code
Hi

Referring to this [page](https://specs.amwa.tv/is-04/releases/v1.3.1/docs/4.1._Behaviour_-_Registration.html).

Can a Node encounter the HTTP 500 error on both Registration and Heartbeats or only on Heartbeats?

If a Node can encounter the code on both, should the same logic apply to both Heartbeats and Registration?

Thanks in advance.

Victor.

garethsb:
Resolves #169.


peterbrightwell (approved):
Looks good following the recent tweaks

PR #190, opened 2022-08-01
Correct ancestor to descendant

garethsb:
I'm also concerned that other facets of ancestry queries are not clear:

* If `query.ancestry_type` is omitted, is `400` (Bad Request) returned?
* When `query.ancestry_generations` is omitted, does that mean _all_ ancestors/descendants, or up to a default value decided by the server (the latter would match `query.paging_limit` behaviour)?

I also note the inconsistency that when `query.ancestry_generations` is greater than supported, `400` (Bad Request) seems to be required, whereas when `query.paging_limit` is greater than supported, the server just indicates the limit used with `X-Paging-Limit`. Why don't ancestry queries behave the same using `X-Ancestry-Generations`?

### DIFF
--- a/docs/APIs - Query Parameters.md
+++ b/docs/APIs - Query Parameters.md
@@ -599,13 +599,13 @@ X-Ancestry-Generations: 4
 
 Payload
 
-- Returns Sources which list an ID of `c1398579-15bc-468e-91ec-df5bbefe1cd3` in their `parents` attribute array. These are the first generation ancestors.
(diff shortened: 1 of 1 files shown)


andrewbonney (approved):

peterbrightwell (approved):
"SHOULD NOT" -> "do not"...    That's probably ok as the former isn't very helpful in practice.  Maybe needs to be MUST NOT in a future normative release.

cristian-recoseanu (approved):

PR #197, opened 2023-03-31
Tweak untestable conformance language

garethsb:


alabou (approved):

docs/Controllers.md
```diff
@@ -82,7 +82,7 @@ In order to avoid polling of the HTTP API, it is RECOMMENDED that Controllers us
 
 ## Dynamic Update of Resources
 
-The Controller MUST be capable of using the Query API to discover and dynamically report the state of any registered resource, including Node, Device, Source, Flow, Sender, and Receiver.
+The Controller uses the Query API to discover and track the state of any registered resource, including Node, Device, Source, Flow, Sender, and Receiver.
```
N-Nagorny: Should we mention that there is no Query API in the Peer to Peer mode and give some guidance on what to do in this case?

garethsb: Pretty much everything else in this Controller spec assumes use of the Query API. I don't disagree that advice about what to do in the absence of a Registry is a good idea, but that's a separate topic/PR it seems to me.

garethsb: On second thoughts, the possibility of operating in peer-to-peer mode could be why the spec was written that way, although no-one remembered that when I asked. 🙂

I realise there's another instance of "The Controller MUST be capable of..." a bit earlier in the document too.

Perhaps separating it into two sentences, i.e. "The Controller MUST be capable of operating in registered mode." followed by adding "In registered operation,..." at the start of proposed change is a better idea?

N-Nagorny: @garethsb, yes, should be very clear for readers.

garethsb: Updated, ready for review again.

docs/Controllers.md
```diff
@@ -65,7 +65,7 @@ Controllers MUST observe and interpret all of the TXT records returned with the
 
 ## Query API
 
-The Controller MUST be capable of using the Query API to discover any registered resource, including Node, Device, Source, Flow, Sender, and Receiver,
+In registered operation, the Controller MUST use the Query API to discover any registered resource, including Node, Device, Source, Flow, Sender, and Receiver,
```
alabou: Is there a definition of "registered operation" ?

N-Nagorny: It's covered in `Discovery` section. It would be good to capitalize "Registered Operation" everywhere in this PR.

garethsb: Done, thanks!

peterbrightwell (approved):
Looks to improve language

PR #198, opened 2023-05-12
Clarify language around Tags and add link to the Tags register in the NMOS Parameter Registers

garethsb:
Cf. AMWA-TV/is-13#10.


peterbrightwell (approved):
All good

garethsb:
Reviewed in ARG call today

PR #200, opened 2023-05-17
Move tags section adjacent to id, version, label and description reflecting definition in resource_core.json schema

garethsb:


peterbrightwell (approved):
Looks good

PR #216, opened 2024-11-21
Changed wording

TheSaneWonko:
Changed wording to allow non-IANA media types which are otherwise well defined

### DIFF
--- a/APIs/schemas/flow_audio_coded.json
+++ b/APIs/schemas/flow_audio_coded.json
@@ -12,7 +12,7 @@
       ],
       "properties": {
         "media_type": {
-          "description": "Subclassification of the format using IANA assigned media types",
+          "description": "Subclassification of the format using IANA assigned media types, if available",
           "type": "string",
           "pattern": "^audio\\/[^\\s\\/]+$",
           "not": {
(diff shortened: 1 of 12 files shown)


peterbrightwell (approved):
Looks good

APIs/schemas/flow_audio_coded.json
```diff
@@ -12,7 +12,7 @@
       ],
       "properties": {
         "media_type": {
-          "description": "Subclassification of the format using IANA assigned media types",
+          "description": "Subclassification of the format using IANA assigned media types, if available",
```
cristian-recoseanu, line 15: Should we try to write these in such a way so it doesn't seem that the field is optional if there isn't an IANA assigned media type?

```suggestion
          "description": "Subclassification of the format (prefer IANA assigned media types where available)",
```

APIs/schemas/receiver_audio.json
```diff
@@ -25,7 +25,7 @@
           "type": "object",
           "properties": {
             "media_types": {
-              "description": "Subclassification of the formats accepted using IANA assigned media types",
+              "description": "Subclassification of the formats accepted using IANA assigned media types, if available",
```
cristian-recoseanu, line 28: Should we try to write these in such a way so it doesn't seem that the field is optional if there isn't an IANA assigned media type?

```suggestion
              "description": "Subclassification of the formats (prefer IANA assigned media types where available)",
```

PR #217, opened 2024-12-04
Changed wording

TheSaneWonko:
Changed wording to allow non-IANA media types which are defined in the NMOS Parameter Registers, as discussed in today's ARG meeting.


garethsb (approved):
LGTM

